PR #71, opened 2014-12-29
Onomy addition

c0cky:
code to accomplish 3 goals

1. Allow for an arbitrary amount of 'concepts' or 'vocabularies' to be created. Made the maximum 100 instead of the previous maximum which was 3.
2. Alter the UI for an input text box and two buttons, one for submission and one for refreshing. I had to add an image for the refresh. The submit grabs the onomy url from the input box and populates the vocabularies and terms from that particular url with a json extension ex: onomy.org/published/50/json. The refresh button grabs the urls from vocabularies that were previously created or had entries from the onomy submit button and 'refreshes' the vocabulary and terms with new additions from the onomy url. (if any) I had to edit the database model of vocabulary to save the url that a user entered. For now this is just a csv serialization of the urls. From my understanding there should not be but a couple if more than one url for a particulary vocabulary so this should not max out this field.
3. Create an export function in a format suited for the MEP using lxml etree. Exports the annotations and information about collected videos by the particular user calling on the url. The url for this export is 'projecturlhere/asset/MEPdump'


mediathread/settings_shared.py
```diff
@@ -23,17 +23,15 @@
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': 'mediathread',
+        'NAME': 'camron',
         'HOST': '',
-        'PORT': '',
-        'USER': '',
-        'PASSWORD': '',
+        'PORT': '5432',
```
sdreher: To customize your local environment, add a local_settings.py in the <app>/mediathread directory. (Alongside the settings_shared.py file). Settings in local_settings.py automatically override the default values in settings_shared.py. local_settings.py is not checked into the repository. Making changes directly to settings_shared.py (as here) can globally affect other installations.

media/js/app/taxonomy/taxonomy.js
```diff
@@ -1,19 +1,19 @@
...
-        }        
-    });    
+        }
+    });
 
     var TermList = Backbone.Collection.extend({
         urlRoot: '/api/term/',
         model: Term,
-        comparator: function(obj) {
+        comparator: function (obj) {
```
sdreher: A minor note: Our team here tries to adhere to community standards for our python and javascript code. (See here http://javascript.crockford.com/code.html). The standard here is not to follow the "function" declaration here with a space.

mediathread/taxonomy/models.py
```diff
@@ -44,6 +43,24 @@ class VocabularyForm(forms.ModelForm):
     class Meta:
         model = Vocabulary
 
+#class Onomy(models.Model):
```
sdreher: Ideally, commented code is removed before integration into the master branch.

mediathread/taxonomy/models.py
```diff
@@ -21,12 +21,11 @@ class Vocabulary(models.Model):
     display_name = models.CharField(max_length=50)
     description = models.TextField(null=True, blank=True)
     single_select = models.BooleanField(default=False)
-
+    onomy_url = models.TextField(null=True, blank=True)
```
sdreher: I'm not seeing the schema migration file for this field addition. Can you verify it has been added to this pull request?

sdreher:
Merged code into new branch onomy.

PR #102, opened 2015-03-18
:arrow_up: pylint 1.4.1

thraxil:


mediathread/settings_shared.py
```diff
@@ -44,8 +44,6 @@
     }
 
 JENKINS_TASKS = (
-    'django_jenkins.tasks.run_pylint',
```
nikolas: Why were these jenkins tasks removed here?

PR #117, opened 2015-04-02
Django 1.7 upgrade

sdreher:


thraxil:
On the staging deploy, it looks like there was a migration issue with djangosherd. Fixed by faking out the migration that appeared to already be applied: 

    ./manage.py migrate djangosherd 0002 --fake --settings=mediathread.settings_staging

Imagine you'll have to do the same on prod.

PR #122, opened 2015-04-07
Update jQuery to v1.11.2, jQuery UI to v1.11.4

nikolas:


sdreher:
I'd like to do some local testing before I merge this one.

sdreher:
I've verified that this upgrade is causing issues in the interface. We'll need to wait to merge this until they can be resolved.

nikolas:
Cool, thanks for looking into this. I'll try running the selenium tests tomorrow and see if I can address the problems.

nikolas:
It looks like updating jQuery UI fixed the lettuce tests, so this PR can be reviewed again.

I've also updated some test dependencies, removing local versions from `requirements/src`.

I'm including flowplayer with the `static` templatetag instead of `http://ccnmtl.columbia.edu/remote/flowplayer-5.5.0/flowplayer.min.js`. Including this js file with `http://` isn't possible when using https, which is how I'm accessing mediathread locally.

sdreher:
Very cool. I'm going to do a little local testing then merge.

PR #130, opened 2015-04-23
django-markwhat 1.3

sdreher:
upgrade to django-markwhat-1.3. previous version errored on pypi


nikolas:
It was partially my fault. https://github.com/Alir3z4/django-markwhat/issues/21
django-markwhat was changing version number schemes which was confusing, and one solution was to remove the date-based version scheme from pypi.

PR #142, opened 2015-05-13
Refactor collaboration: Replace Policy Singleton with ObjectManager

sdreher:


structuredcollaboration/models.py
```diff
@@ -153,3 +162,46 @@ def set_policy(self, p):
     def __unicode__(self):
         return u'%s %r <%s %s> [%s]' % (self.title, self.pk, self.content_type,
                                         self.object_pk, self.slug)
+
+
+CollaborationPolicyRecord.objects.register_policy(
+    CollaborationPolicy, 'forbidden', 'Forbidden to everyone')
+
```
thraxil: I know it's worked this way for quite a while, but I'm not really a fan of things registering themselves as a side-effect of being imported.

With Django 1.7+, we ought to be able to at least move these into an `AppConfig.ready()`

sdreher: Agreed. I'll take a look at AppConfig.ready and think through an alternate approach.

PR #160, opened 2015-05-29
Taxonomy refactor

c0cky:
*Removed a bunch of inline functions 
*Removed hard to read lines
*Made use of Vocabulary.addTerm() to speed up code.


sdreher:
The build failed due to some stray jshint errors. Click the "Details" to the right of the "Failed -- The Travis CI build failed" for review the errors. When you checkin the fixes to your master branch, the changes will be picked up here, and will rerun the Travis build. 

A good habit to foster is to always run unit tests and code validation steps before a code checkin. In our case, that's the unit test suite as well as django validate, flake8 and jshint. All tests/checks are detailed in the ./Makefile. Simply execute "make" in the root mediathread directory to run these basic tests.

To go one step further, you can also run the selenium tests via the "make harvest" command. This suite takes about 25 minutes to complete. I'll usually run a harvest before I issue a pull request, as opposed to every time I check in code.

media/js/app/taxonomy/taxonomy.js
```diff
@@ -486,110 +488,42 @@
...
-                                                    self.render();
-                                                }
-                                            });
-                                        }
-                                    }
-                                }(j));
-                            }
-                        }
                     } else if (display !== undefined && display.length > 0) {
                         var urls = selectedVocabulary.getOnomyUrls();
```
sdreher: In this section, sectionVocabulary is potentially saved multiple times. Could selectedVocabulary simply be modified here, then saved with the newly created variables at the end?

media/js/app/taxonomy/taxonomy.js
```diff
@@ -612,10 +546,48 @@
                             });
                         }
                     }
+                    //once we get to last iteration we save it all
+                     if (i == MAX - 1) {
```
sdreher: As a final save step, this section is probably better placed outside of the "onomy data" for loop.

media/js/app/taxonomy/taxonomy.js
```diff
@@ -486,110 +488,42 @@
             var self = this;
 
             jQuery.get(onomyURL, function(data) {
-                var x = data;
-
                 var parents = [];
```
sdreher: Could the parents array be modeled as a map, indexed by display_name? This would allow easy lookup and mitigate the need for the inline search.

media/js/app/taxonomy/taxonomy.js
c0cky: Could you please elaborate on how you think it could be saved multiple times? My intention is to add the onomy_url to the vocabulary if it doesn't already contain it then, if the term doesn't exist add it to the vocabulary.

media/js/app/taxonomy/taxonomy.js
c0cky: Wow thanks for that tip, I didn't know variables were in the same scope as the for loop they're declared in for javascript.

media/js/app/taxonomy/taxonomy.js
sdreher: Sure. As written, the selectedVocabulary should be saved due to a url addition or term addition/change. My thought was to have selectedVocabulary just be updated in this section -- urls and terms -- then saved once after the complete iteration of the onomy payload. (Ideally, I'd like to get rid of the complexity of the queuedSave completely.) Would this approach work?

media/js/app/taxonomy/taxonomy.js
sdreher: The syntax here should be parents = {};

sdreher:
Really excellent refactor. I'm going to merge as it works -- and it works so much faster! I'll tweak the parents = {} on this side. Thank you!

PR #170, opened 2015-06-03
Convert homepage.mustache to standard mustache - PMT #96898

nikolas:
This is the first step towards using standard Mustache.js instead
of our custom CCNMTL version.

* I've removed all the unnecessary files from the media/js/mustache
  directory.
* I converted the homepage.mustache template to standard mustache. This
  required some code changes in projectlist.js.
* There are two versions of Mustache working on the site in this branch:
  Standard Mustache 2.0.0, and our existing version of
  customized Mustache 0.4.0. I'll convert over the other templates in
  future pull requests, and eventually we can get rid of the old mustache.


nikolas:
I can point this pull request to the fall2015 branch if that's better

sdreher:
Yes. That would be better.

PR #173, opened 2015-06-03
Use on/off instead of bind/unbind - PMT #101482

nikolas:


media/js/app/assetmgr/collection.js
```diff
@@ -482,7 +482,7 @@ CollectionList.prototype.updateSwitcher = function () {
     Mustache.update("switcher_collection_chooser", self.switcher_context, { parent: self.parent });
 
     // hook up switcher choice owner behavior
-    jQuery(self.el).find("a.switcher-choice.owner").unbind('click').click(function (evt) {
+    jQuery(self.el).find("a.switcher-choice.owner").off('click').on('click', function (evt) {
```
sdreher: The bind/unbind behavior was originally written as the client-side template/dom element replacement caused issues. A handler attached to a non-existent element would fire incorrectly. Theoretically, the unbind might be eliminated now and replaced with a .on statement following a delegated event pattern.

nikolas: Yeah thanks for pointing that out .. I will update this.

nikolas: When I switched this to a delegate pattern, e.g.:

```js
jQuery(self.el).on('click', 'a.switcher-choice.owner', function (evt) {
```
there were some lettuce test failures. I switched it back to `.off().on()` for now, because in this state it's working. I'm okay with refactoring this event listener to a delegate pattern later, but if you think it should be addressed now I'm also fine with looking deeper into the problem.

sdreher: Thanks for experimenting. When we have some time, it might be interesting to dig deeper into this. But I'm good going with this now.

PR #175, opened 2015-06-05
Convert homepage.mustache to standard mustache - PMT #96898

nikolas:
**Re-opening PR #170 against the fall2015 branch.**

This is the first step towards using standard Mustache.js instead
of our custom CCNMTL version.

* I've removed all the unnecessary files from the media/js/mustache
  directory.
* I converted the homepage.mustache template to standard mustache. This
  required some code changes in projectlist.js.
* There are two versions of Mustache working on the site in this branch:
  Standard Mustache 2.0.0, and our existing version of
  customized Mustache 0.4.0. I'll convert over the other templates in
  future commits, and eventually we can get rid of the old mustache.


media/templates/homepage.mustache
```diff
@@ -1,236 +1,256 @@
...
-                            <a class="switcher-choice owner" href="{{space_viewer.username}}" style="font-weight: bold">Me</a>
+<div>
+  <div class="filter-widget left">
+    {{#switcher_collection_chooser}}
+      <div class="switcher_collection_chooser switcher collection-chooser" tabindex="1">
+        <a class="switcher-top menuclosed" onclick="return switcher(event, this);">
+          <span class="switcher-prefix">Author </span>
+          <span class="title">
+            {{#selected_label}}
+              {{{selected_label}}}
```
sdreher: The ellipses filter truncates the string to a certain length, then adds an ellipses. Is this functionality lost under Mustache2?

nikolas: Mustache allows you to do template functions like this by passing them into the context:
https://github.com/janl/mustache.js#functions

I'll add an ellipsis function so we retain this functionality.

PR #190, opened 2015-06-30
Enhanced/downloadable assignment report

sdreher:


mediathread/djangosherd/models.py
```diff
@@ -118,8 +118,8 @@ def filter_by_date(self, date_range):
         return self.filter(Q(added__range=[startdate, enddate]) |
                            Q(modified__range=[startdate, enddate]))
 
-    def get_related_notes(self, assets, record_owner, visible_authors,
-                          all_items_are_visible,
+    def get_related_notes(self, assets, record_owner, visible_authors=[],
```
thraxil: Danger, Will Robinson!  Mutable default arguments!

sdreher: ah, good catch. i will resolve.

PR #196, opened 2015-07-09
Mustache 2 conversion

nikolas:
This started as an attempt to convert the project template to standard Mustache, but
through the lettuce failures I realized I would need to convert over all the Mustache
stuff in the asset viewer as well. This is mostly in `asset.js`.

I've split out the `asset_workspace` template into different pieces that we can
render individually with `Mustache.render`. This was necessary because we're
not using `Mustache.update` anymore.

Most of the work is now done, there's just a few more places in `collection.js` and
`projectpanel.js` that use the old Mustache, which I'll address in a different PR, after
this step is complete.


nikolas:
Here's some more in depth notes on this change: http://www.columbia.edu/~njn2118/journal/2015/7/10.html

media/js/app/assetmgr/asset.js
```diff
@@ -606,24 +643,27 @@
...
-                                        self._initConcepts();
-                                        self._initReferences();
-                                        jQuery(window).trigger('resize');
-                                    });
+                        jQuery('.asset-view-title').text(
+                            context['asset-current'].title);
+
+                        var $elt = jQuery('#asset-global-annotation');
+                        $elt.hide();
+                        rendered = Mustache2.render(
```
sdreher: need to declare this variable.

sdreher:
Nik - Sorry for the delay on the review. I'm working through this now.

nikolas:
it's cool.. no hurry. I have plenty of other things to work on!

PR #209, opened 2015-07-18
:bug: Class references not updating for projects

sdreher:
As part of the big collaboration refactoring, the class reference indexer was broken due to a reordering of events.

This fixes the issue by:
1. removing the ancient and rather inelegant universal post_save hook.
2. Calling explicit update_class_reference functions.

Ideally, the whole DiscussionIndex model would be factored out completely. But, one step at a time.

Selenium tests on another branch were added to demonstrate this issue (https://github.com/ccnmtl/mediathread/pull/206)

### DIFF
--- a/mediathread/discussions/utils.py
+++ b/mediathread/discussions/utils.py
@@ -1,7 +1,9 @@
 from courseaffils.models import Course
 from django.contrib.contenttypes.models import ContentType
-from structuredcollaboration.models import Collaboration
 from threadedcomments.models import ThreadedComment
+
+from mediathread.djangosherd.models import SherdNote, DiscussionIndex
+from structuredcollaboration.models import Collaboration
 
 
 def get_course_discussions(course):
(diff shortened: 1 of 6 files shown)


mediathread/discussions/utils.py
```diff
@@ -49,3 +51,25 @@ def pretty_date(timestamp):
...
+
+    for note in notes:
+        try:
+            disc, created = DiscussionIndex.objects.get_or_create(
+                participant=comment.user,
+                collaboration=comment.content_object,
+                asset=note.asset)
+            disc.comment = comment
+            disc.save()
+        except:
```
thraxil: Can we catch the specific exception instead of having a bare except?

mediathread/projects/models.py
```diff
@@ -407,6 +408,19 @@ def citations(self):
         note_model = models.get_model('djangosherd', 'SherdNote')
         return note_model.objects.references_in_string(self.body, self.author)
 
+    def update_class_references(self):
```
thraxil: This function is *almost* the same as the one in `mediathread/discussions/utils.py`. Seems like we ought to be able to factor out the common functionality.

sdreher: Sure. I was quickly moving some old code around. Not even sure what the "# some things may be deleted?" comment references on that bare exceptions. I'll make a second pass.

sdreher:
This is ready for re-review.

PR #228, opened 2015-07-28
add skos_uri

c0cky:
*add skos_uri to taxonomy model and migrations for the changes.

only changes are adding these two fields. Travis CI should not fail.


sdreher:
@c0cky - the selenium tests were failing as the base sqlite3 test database (scripts/lettuce_base.db) had not been migrated. I updated that & merged the PR.

PR #230, opened 2015-07-29
Replace search box with bootstrap input-group + button

sdreher:
& don't autosubmit the project save form

### DIFF
--- a/media/templates/collection.mustache
+++ b/media/templates/collection.mustache
@@ -7,7 +7,7 @@
                 <div class="switcher-tool switcher-search-text input-group">
                     <span class="input-group-addon"><span class="glyphicon glyphicon-search"></span></span>
                     <input type="text" class="form-control" name="search-text"
-                        value="{{active_filters.search_text}}" placeholder="Item & selection titles"/>
+                        value="{{active_filters.search_text}}" placeholder="Item &amp; selection titles"/>
                     <span class="input-group-btn">
                         <button class="btn btn-default btn-search-text" type="button">Go</button>
                     </span>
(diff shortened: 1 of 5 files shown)


media/templates/collection.mustache
```diff
@@ -2,15 +2,18 @@
...
-                <div class="switcher-tool switcher-search-text {{#active_filters.search_text?}}populated{{/active_filters.search_text?}}">
-                    <span class="search-text-prefix"><img src="/media/img/icons/search.png" /></span>
-                    <input type="text" name="search-text" value="{{active_filters.search_text}}" placeholder="Item and selection titles"/>
-                    <span class="search-text-clear"><img src="/media/img/icons/meth_clear.png" /></span>
+            <div class="switcher-search-container">
+                <h3>SEARCH</h3>
+                <div class="switcher-tool switcher-search-text input-group">
+                    <span class="input-group-addon"><span class="glyphicon glyphicon-search"></span></span>
+                    <input type="text" class="form-control" name="search-text"
+                        value="{{active_filters.search_text}}" placeholder="Item & selection titles"/>
```
thraxil: Does the ampersand need to be encoded here (and in gallery.mustache)? Or is mustache smart enough to do the right thing?

sdreher: Good question. Mustache is okay with the unescaped ampersand. Let me test with the &amp; though and see what happens.

sdreher: Seems like escaped or unescaped is fine. I'll update it with the `&amp;` as that is more correct.

PR #234, opened 2015-07-31
Fall2015

c0cky:
*add skos import into getTheOnomy
*save skos_uri to Term and Vocabulary Models


media/js/app/taxonomy/taxonomy.js
```diff
@@ -474,16 +494,47 @@
                 return item.display_name == thing;
             });
         },
-        getTheOnomy: function(onomyURL, selectedVocabulary) {
+        getTheOnomy: function(onomyURL, selectedVocabulary, JSON_FLAG) {
             var self = this;
 
             jQuery.get(onomyURL, function(data) {
+                console.log(data);
```
sdreher: please remove console.logs

media/js/app/taxonomy/taxonomy.js
```diff
@@ -474,16 +494,47 @@
                 return item.display_name == thing;
             });
         },
-        getTheOnomy: function(onomyURL, selectedVocabulary) {
+        getTheOnomy: function(onomyURL, selectedVocabulary, JSON_FLAG) {
```
sdreher: regular javascript variable names should be described in camelCase. (though we are guilty of mixing in the python underbar syntax!)  This all uppercase variable name might be confused with a global constant.

media/js/app/taxonomy/taxonomy.js
```diff
@@ -442,31 +448,45 @@
 
             // split the url.
             var urls = value.split(',');
+            var flagArray = [];
```
sdreher: Rather than messing around with this flagArray, I'd opt for doing the regex check twice -- once for the valid url check at line 462 and once for the getTheOnomy call at line 477.  Simplifies the code.

media/js/app/taxonomy/taxonomy.js
```diff
@@ -474,16 +494,47 @@
...
+
+                        try {
+                            pL = data[skos_uri]["http:\/\/onomy.org\/onomy-ns#parentLabel"].value.trim();
+                        } catch (e) {
+                            pL = undefined;
+                        }
+                    }
 
-                    if (typeof pL !== undefined && pL.length > 0) {
+                    console.log(display);
```
sdreher: please remove console.logs

media/js/app/taxonomy/taxonomy.js
```diff
@@ -75,6 +75,12 @@
                 var term = new Term({display_name: termName});
                 this.get('term_set').add(term);
             }
+        },
+        addTermWithURI: function(termName, uri) {
```
sdreher: The base addTerm function at line 73 is no longer referenced. Rather than add this new function, a simple refactor on the original addTerm would be more efficient.

media/js/app/taxonomy/taxonomy.js
sdreher: The addition of the skos logic to the getTheOnomy function is causing it to grow again. What do you think of splitting the logic and having a skos importer and  a json importer with some helper functions?

media/js/app/taxonomy/taxonomy.js
c0cky: What do you recommend for the testing case if I remove the flagArray?

sdreher: Hmmm, I'd probably create two helper functions -- isJson(url), isSkos(url) that returns true or false based on the regex match. Then, at Ln. 462, if (!isJson(urls[i]) && !isSkos(urls[i]) { show error }. At line  477,  this.getTheOnomy(urls[j], this.selected, isJson(urls[j]); The helper function can also be used in refreshOnomy to maintain consistency rather than the quick (urls[i].indexOf("json") > -1) check.

c0cky: I will have to add the (urls[i].indexOf("json") > -1) check when calling getTheOnomy at line 477 because the test url will not match the regex.

sdreher: Good point. Maybe the isJson regex could be expanded to check for either the onomy url pattern or the test url pattern.

media/js/app/taxonomy/taxonomy.js
```diff
@@ -434,6 +434,26 @@
             });
             return false;
         },
+        isSKOS: function(url) {
+            var skos_regex = /onomy.org\/published\/(\d+)\/skos/g;
+            var skos_match = skos_regex.exec(url);
+            if (skos_match !== null) {
```
sdreher: nit: return skos_match !== null

sdreher:
Good refactor & much cleaner now. I like how the big getTheOnomy function was broken down.

PR #236, opened 2015-08-03
fix undefined var error

sdreher:
This showed up in my local env, but was not caught by Travis. Odd.

> make jenkins
./manage.py check
System check identified no issues (0 silenced).
./node_modules/jshint/bin/jshint media/js/app/
media/js/app/taxonomy/taxonomy.js: line 567, col 32, 'skosData' is not defined.
1 error
make: *** [jshint] Error 2


c0cky:
That is strange. Didn't show up on my 'make jshint' either. Good catch.

PR #249, opened 2015-08-07
Consolide Project Views

sdreher:
**ProjectWorkspaceView + SelectionAssignmentView**
Initially, my thought was to separate out the SelectionAssignmentView into its own url/view. This proved to be somewhat problematic as MeTh feels projects (assignments, compositions, responses) live at a certain place. And, hacking out the Project.get_absolute_url function began to get messy. This PR unites the project view via a common CBV.

**SelectionAssignment creation**
SelectionAssignments are not created until valid properties are in place. The creation of empty projects is a MeTh anti-pattern that has caused much confusion.)

### DIFF
--- a/mediathread/mixins.py
+++ b/mediathread/mixins.py
@@ -185,8 +185,7 @@ class ProjectVisibleMixin(object):
 class ProjectEditableMixin(object):
     def dispatch(self, *args, **kwargs):
         project = get_object_or_404(Project, pk=kwargs.get('project_id', None))
-        if (project and
-                not project.can_edit(self.request.course, self.request.user)):
+        if (not project.can_edit(self.request.course, self.request.user)):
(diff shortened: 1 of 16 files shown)


mediathread/mixins.py
```diff
@@ -185,7 +185,8 @@ def dispatch(self, *args, **kwargs):
 class ProjectEditableMixin(object):
     def dispatch(self, *args, **kwargs):
         project = get_object_or_404(Project, pk=kwargs.get('project_id', None))
-        if not project.can_edit(self.request.course, self.request.user):
+        if (project and
```
thraxil: With a `get_object_or404()` on the previous line, the extra check here seems redundant.

sdreher: quite. I'll resolve.

PR #260, opened 2015-08-12
Selection Assignment View: associate notes with assignment response

sdreher:
* ProjectNote - create an relationship between a project and an annotation on annotation creation.
* ProjectItemView + RestrictedMaterialsMixin: display notes based on assignment policies and response submit state.
* tests
* refactoring/renaming as needed


mediathread/factories.py
```diff
@@ -151,6 +151,14 @@ class CollaborationPolicyRecordFactory(factory.DjangoModelFactory):
     FACTORY_FOR = CollaborationPolicyRecord
 
 
+class AssignmentItemFactory(factory.DjangoModelFactory):
+    FACTORY_FOR = AssignmentItem
```
thraxil: Heads up that `FACTORY_FOR` is deprecated in factoryboy now, so we should be switching to

```
class Meta:
    model = AssignmentItem
```

style instead.

PR #267, opened 2015-08-13
Fix AttributeError when checking permission

nikolas:
After pulling the latest fall2015 changes, I was getting
this error when loading the homepage:

> AttributeError at /api/project/user/ccnmtl/
> 'NoneType' object has no attribute 'permission_to'
>
> 450.         if not collaboration.permission_to('read', course,
> viewer):
>
> (in mediathread/projects/models.py)

This change just checks to see if `collaboration` is `None` before
calling `permission_to` on it.

I'm not sure if this behavior is correct - maybe we should be returning
`True` when `collaboration` is `None`?

### DIFF
--- a/mediathread/projects/models.py
+++ b/mediathread/projects/models.py
@@ -447,7 +447,8 @@ class Project(models.Model):
     def can_read(self, course, viewer):
         # has the author published his work?
         collaboration = self.get_collaboration()
-        if not collaboration.permission_to('read', course, viewer):
+        if (collaboration is None) or \
+           (not collaboration.permission_to('read', course, viewer)):
             return False
 
         # assignment response?


sdreher:
Seems odd that a project was created without a collaboration...though it might be legacy stuff floating around. Do you know the id of the project it blew up on?

nikolas:
Yeah I figured it must be some stale data or something.

The project ID is 15, the title is: "Bel's Test  Project".
* course_id: 1
* author_id: 14

sdreher:
heh - project id 15 was created 2010-01-26. I think returning False from can_read is reasonable. I'll setup a task to deal with empty project collaborations so we won't have to code around a mess going forward.

PR #268, opened 2015-08-13
Selection View Tweaks

sdreher:
Selection Assignment Create/Save
* body was not saving on create
* due_date save/display bug
* reorder wizard pages
* remove public to world as a save option

Selection Assignment View
* display 'no responses' to instructor when applicable

### DIFF
--- a/media/js/app/projects/selection_assignment_edit.js
+++ b/media/js/app/projects/selection_assignment_edit.js
@@ -1,4 +1,4 @@
-/* global _: true, Backbone: true */
+/* global _: true, Backbone: true, CitationView: true */
 /* global showMessage: true, tinyMCE: true */
 
 (function(jQuery) {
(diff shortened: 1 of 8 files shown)


media/js/app/projects/selection_assignment_edit.js
```diff
@@ -36,23 +35,32 @@
...
         validate: function(pageNo) {
             var q;
             if (pageNo === 1) {
-                return jQuery(this.el).find('input[name="title"]')
-                                      .val()
-                                      .length > 0;
+                return jQuery('input[name="item"]').val() !== '';
             } else if (pageNo === 2) {
-                return tinyMCE.activeEditor.getContent().length > 0;
+                var title = jQuery(this.el).find('input[name="title"]').val();
```
nikolas: just a minor thing, but you can use `this.$el` instead of initializing `jQuery(this.el)` each time. http://backbonejs.org/#View-$el

sdreher: good tip. thx.

PR #270, opened 2015-08-13
Enable jsonp response for is_logged_in view

nikolas:
To get around cross-origin problems with the chrome extension


nikolas:
I'm closing this because I've been having some trouble getting JSONP to work with the chrome extension, but using CORS headers with the normal JSON response seems to work, so I'll make a PR for that.

PR #275, opened 2015-08-14
Fix icon alignment (see screenshots)

nikolas:
The mis-alignment was probably caused by the bootstrap upgrade.

This fixes the alignment of the question-mark icon on the homepage.

before:
2015-08-14-151329_605x168_scrot

after:
2015-08-14-151337_588x166_scrot

### DIFF
--- a/media/css/mediathread.css
+++ b/media/css/mediathread.css
@@ -2509,7 +2509,6 @@ div.gallery-item h4.asset_title {
 
 .actions {
     float: right;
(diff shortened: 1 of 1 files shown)


sdreher:
the .actions class is used in a few places in the app. can you verify this looks okay in the other locations or scope the style change down to the homepage? (note: check both the django and mustache templates for references)

nikolas:
postponing until next release

PR #287, opened 2015-08-19
Check that jQuery element exists before cloning - PMT #102478

nikolas:
This fixes the JS error that occurs when the target element
that switches for the Selection Assignment doesn't exist (like
when the window is too narrow).

### DIFF
--- a/media/js/app/projects/selection_assignment_edit.js
+++ b/media/js/app/projects/selection_assignment_edit.js
@@ -155,6 +155,10 @@
             $overlay.addClass('hidden');
         },
         onGalleryItemSelect: function(evt) {
+            if (!this.hoverItem) {
+                return;
+            }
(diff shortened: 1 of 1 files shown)


media/js/app/projects/selection_assignment_edit.js
```diff
@@ -158,11 +158,13 @@
             var pk = jQuery(this.hoverItem).attr('data-id');
```
sdreher: this.hoverItem is referenced above the check.

Theoretically, the user shouldn't be able to select something when there is no hoverItem. Unless there's a race condition between mouseOut & the selection.

nikolas: I update this code to just return if this.hoverItem doesn't exist.

PR #288, opened 2015-08-19
selection assignment tweaks

sdreher:
* language updates
* image placeholder

### DIFF
--- a/mediathread/templates/projects/selection_assignment_edit.html
+++ b/mediathread/templates/projects/selection_assignment_edit.html
@@ -88,19 +88,20 @@
                          shared with the entire class or just the instructor.
                         </p>
                         <ol>
-                            <li>Choose an item in your collection.</li>
-                            <li>Write Instructions</li>
-                            <li>Set Response Due Date & Visibility</li>
-                            <li>Publish Assignment</li>
+                            <li>Choose an item from your collection</li>
+                            <li>Write instructions</li>
+                            <li>Set response due date & visibility</li>
+                            <li>Publish assignment</li>
                         </ol>
-                        <ul class="pager">
-                            <li><a href="#" class="left prev">Previous</a></li>
-                            <li><a href="#" class="right next">Next</a></li>
-                        </ul>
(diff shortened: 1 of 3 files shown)


mediathread/templates/projects/selection_assignment_edit.html
```diff
@@ -88,19 +88,20 @@
...
+                        <br />
+                        <div class="text-center">
+                        <a href="#" class="btn btn-primary next right">Get Started</a>
+                        </div>
                     </div>                   
                     <div data-page="2" class="page hidden">
-                        <h3>1. Choose an item from the course collection</h3>
+                        <h3>1. Choose an item from your collection</h3>
                         <div class="selected-item">
+                        <center>
```
nikolas: I think you should use a div with CSS set to `text-align: center;` here, because the `<center>` element has been deprecated in html5.

* http://www.w3schools.com/tags/tag_center.asp
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center

mediathread/templates/projects/selection_assignment_edit.html
```diff
@@ -88,19 +88,20 @@
                          shared with the entire class or just the instructor.
                         </p>
                         <ol>
-                            <li>Choose an item in your collection.</li>
-                            <li>Write Instructions</li>
-                            <li>Set Response Due Date & Visibility</li>
-                            <li>Publish Assignment</li>
+                            <li>Choose an item from your collection</li>
+                            <li>Write instructions</li>
+                            <li>Set response due date & visibility</li>
```
nikolas: I think this ampersand should be encoded as `&amp;`

mediathread/templates/projects/selection_assignment_edit.html
sdreher: heh - you caught me using old tags. I'll update.

sdreher:
Closing as I'm rethinking the whole placeholder thing. Stay tuned.

PR #298, opened 2015-08-21
Selection Assignment Edit View: Fix missing thumbs

sdreher:
Images may/may not have thumbnails. If no thumbnail exists, simply show the image at smaller dimensions


mediathread/templates/projects/selection_assignment_edit.html
```diff
@@ -120,9 +120,15 @@ <h4 class="asset_title asset-{{item.primary.label}}">
...
                                     </h4>
                                     <div class="asset-thumb">
-                                        <img class="gallery-item-thumb"
-                                         src="{{item.thumb_url}}"
-                                         alt="Item Thumbnail">
+                                        {% if item.thumb_url %}
+                                            <img class="gallery-item-thumb"
+                                             src="{{item.thumb_url}}"
+                                             alt="Item Thumbnail">
+                                        {% else %}{% ifequal item.media_type "image" %}
```
nikolas: fyi, i think you can do `{% if item.media_type == "image" %}` now. But this looks good to me.

PR #345, opened 2015-09-25
lti fullname > django first/last name

sdreher:
dealing with name parsing in a more sensible way.


lti_auth/auth.py
```diff
@@ -4,6 +4,7 @@
 from pylti.common import LTIException
 
 from lti_auth.lti import LTI
+from nameparser import HumanName
```
thraxil: nameparser looks handy. probably some other projects that could be useful on too.

PR #350, opened 2015-09-28
LTI config.xml

sdreher:
* config.xml endpoint for LMS that support this method of configuration
* allow config.xml customization via template & settings options.


thraxil:
Ha! Should've guessed that LTI would come along with a bunch of XML...

PR #352, opened 2015-09-28
Mustache collection conversion - PMT #96898

nikolas:
This gets rid of the old Mustache library, so now all the javascript templates are using the same version of Mustache.js: version 2.x.

This affected the gallery view (at the `/asset/` route) and also the collection column in the composition view. I've refactored the javascript that prepares those views.

mediathread_mustache.js is now called mediathread_setup.js, and I've removed some redundant template-loading code to rely on `MediaThread.loadTemplates()`, defined in that file.

That file also contains the common mustache helper functions we can use throughout mediathread.


media/js/app/panel.js
```diff
@@ -57,23 +68,6 @@
             jQuery(self.el).css('height', visible + 'px');
         };
```
sdreher: The assignment.feature, scenario 2 selenium failure is related to the removal of the loadTemplates function. In panel.js at Ln. 261, self.loadTemplates is still referenced in the newPanel function. (https://github.com/ccnmtl/mediathread/blob/mustache-collection-conversion/media/js/app/panel.js#L261). 

newPanel is a cute little way to automagically open an Assignment Response panel when a student clicks "Respond to Assignment" within the assignment space. If this is a pain to fix, replacing this logic with a simple redirect is a completely fine option.

nikolas: Thanks for looking into this. I'm not sure what led me to change that call to `self.loadTemplates(0)` to using `MediaThread.loadTemplates()` - maybe the goal was consolidating redundant code. I did notice the `newPanel` stuff.. anyways, I'm investigating now.

nikolas: yeah.. I had overlooked one use of `panel.loadTemplates()` in that file. After adding back that function, it looks like the tests are getting a little farther.

nikolas: in fact - that fixes the remaining tests

PR #353, opened 2015-09-28
Asset Workspace - removing duplicate logic

sdreher:
The full asset collection view and the asset detail view were both being called on the same page, and delivering the same data. Consolidating the logic to the asset detail view.


sdreher:
waiting on the mustache2 refactor before I move forward with this.

PR #355, opened 2015-10-02
Canvas LTI Editor Integration

sdreher:
* Specify editor embed information in generated config.xml
* Break out LTILoginView into mixin + routing view
* Asset Embed List view - provides a list of embeddable items & selections
* Asset Embed View - provides a view of a single item or selection.
* test, test, test


media/templates/embed.mustache
```diff
@@ -0,0 +1,281 @@
+{{#>>collection_table}}
```
nikolas: Because this is being rendered by `CollectionList`, which I've recently updated to use Mustache 2, this template will need some updates. In particular, using `{{#>>` to denote a partial won't work. I can help with this merge to the new code in the `fall.midterm.2015` branch if you like.

sdreher: I just followed the collection.mustache example and have it working I think. 3ad69bc

media/templates/embed_assets.mustache
```diff
@@ -0,0 +1,169 @@
+{{^assets}}
+    <div class="default-content intro-text feature-help"">
```
nikolas: There's an extra double-quote character at the end of this line

sdreher: fixed!

PR #377, opened 2015-10-22
Add bookmarklet error when scripts fail to load - PMT #103532

nikolas:
This is my first pass at addressing the content-security-policy failures in the bookmarklet. We will talk about this more in the Mediathread meeting.

In particular, I would like to clarify there how we should direct users to the extension from the error message, and whether or not a simple alert popup is sufficient.


sdreher:
Based on our meeting, do you want to add to this PR or open new PRs as the additional code is developed?

nikolas:
I'll add to this PR so i'll close this for now.

PR #378, opened 2015-10-23
Add bookmarklet js load error - PMT #103532

nikolas:
This adds a bookmarklet migration page that users get forwarded to when a Content-Security-Policy error happens in the bookmarklet.

2015-10-23-113756_451x234_scrot

Pressing "OK" forwards the user to this page:
2015-10-23-113738_950x366_scrot

while pressing "Cancel" dismisses it.


mediathread/assetmgr/views.py
```diff
@@ -1046,3 +1046,7 @@ def get(self, request):
         if len(notes) > 0:
             context = {'tags': TagResource().render_related(request, notes)}
         return self.render_to_json_response(context)
+
+
+class BookmarkletMigrationView(TemplateView):
```
sdreher: The upgrade_bookmarklet view is in mediathread.main.views. Probably a good idea to have these two related views in the same place. Agnostic as to whether they stay in main or assetmgr.

nikolas: I put this in assetmgr because that's where all the bookmarklet-related templates are. I'll move the upgrade_bookmarklet view over here.

nikolas:
This is ready for review again

PR #409, opened 2015-11-04
Fix Vimeo mixed-content error

nikolas:
This fixes a mixed content error when using mediathread over
HTTPS and viewing a vimeo asset in the collection:

> requested an insecure plugin resource
> 'http://vimeo.com/moogaloop.swf'. This request has been blocked; the
> content must be served over HTTPS.

The video will continue to work correctly when using mediathread
over HTTP.


sdreher:
Note: at the moment, we encourage users to access Mediathread proper via http://. We have at least one content-provider (ARTStor), who hasn't switched over to https:// yet.

PR #425, opened 2015-11-11
Address collaboration/project performance issues

sdreher:
small gains - 25% fewer queries, 30% quicker to load the ProjectCollectionView for Music Humanities

* cache the course collaboration
* refactor visible by course to retrieve all collaborations at once
* Collaboration queries, select_related context, policy_record
* index CollaborationPolicyRecord.name

### DIFF
--- a/mediathread/main/course_details.py
+++ b/mediathread/main/course_details.py
@@ -2,6 +2,7 @@ from courseaffils.models import Course
 from django.core.cache import cache
 
 from mediathread.assetmgr.models import ExternalCollection
+from structuredcollaboration.models import Collaboration
 
 
 UPLOAD_PERMISSION_KEY = "upload_permission"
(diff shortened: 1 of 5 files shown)


thraxil:
Have you done any individual benchmarks on the changes? Specifically I'm wondering if indexing the field might provide the bulk of the improvements vs caching which tends to bring in the potential for other problems. Ie, does the caching bring enough of a performance improvement on its own to be worth it?

sdreher:
I've been measuring each change and including if the query or timing difference seemed significant. Agreed, it's a bit lazy to group these all together. I'll close this one and PR per change with metrics.

Noting -- I have been avoiding caching for the reasons you mention. The course collaboration caching is luckily fairly low stakes. We have about 50 courses running in a given semester and the cache will probably contain a fraction of the total.

PR #445, opened 2015-11-30
docker setup

thraxil:
test running step of Dockerfile is currently commented out because tests fail
on my machine whether in docker or not.

If you have docker and docker-compose installed, you can do:

```
$ docker pull ccnmtl/mediathread
$ docker-compose run web migrate
$ docker-compose up
```

and you will have a dev runserver instance running on port 8000. It appears that mediathread needs some additional configuration tweaks for that to be usable, but I haven't dug into that yet.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+--index-url https://pypi.ccnmtl.columbia.edu/
 Django==1.8.6
 psycopg2==2.6.1
 #for MySQL: MySQL-python==1.2.5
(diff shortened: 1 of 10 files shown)


requirements.txt
```diff
@@ -18,7 +19,7 @@ testtools==1.8.0  # python-subunit
 pbr>=0.11
 testscenarios==0.5.0  # python-subunit
 python-subunit==1.1.0  # lettuce
-requirements/src/lettuce-0.2.21.ccnmtl.tar.gz
+lettuce==0.2.21.ccnmtl
```
sdreher: Externals are configured to pull from pypi.python.org. The vendored lettuce build & django-courseaffils 2.0.6 need to be available there before we can pull these libraries from the source tree.

sdreher: Okay - I see that you're using the --index-url option now. Thanks for explaining that.

PR #467, opened 2015-12-17
Set SECURE_PROXY_SSL_HEADER

sdreher:
Nginx translates https:// scheme into the X-Forwarded-Protocol header.
Configure Django to use this header to determine if a request is secure.
(This configuration change should resolve an oauth handshake issue in production.)

### DIFF
--- a/mediathread/settings_shared.py
+++ b/mediathread/settings_shared.py
@@ -318,6 +318,8 @@ LTI_TOOL_CONFIGURATION = {
     'embed_tool_id': 'mediathread',
 }
 
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTOCOL', 'https')
+
 # if you add a 'deploy_specific' directory
 # then you can put a settings.py file and templates/ overrides there
 # otherwise, make sure you specify the correct database settings in your


mediathread/settings_shared.py
```diff
@@ -318,6 +318,8 @@ def default_url_processor(url, label=None, request=None):
     'embed_tool_id': 'mediathread',
 }
 
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTOCOL', 'https')
```
nikolas: In the django documentation for this setting, they set the first tuple element to `HTTP_X_FORWARDED_PROTO` instead of `HTTP_X_FORWARDED_PROTOCOL`. I just wanted to make sure that's not a typo.

Searching around I found some random info about an `X-Forwarded-Proto` header in nginx.. is there some place we can see this header when making an HTTP request to mediathread?

sdreher: Oh, good eye there. The salted nginx.conf for Meth (and our other apps) shows X-Forwarded-Protocol being used. I'll do a little digging tomorrow to see if one is preferred over the other.

sdreher: Okay, I've verified via the Django Debug Toolbar/Headers panel on staging that the header being sent through via nginx is: X-Forwarded-Protocol.

From what I can gather, [both X-Forwarded-Protocol and X-Forwarded-Proto](http://stackoverflow.com/questions/16042647/whats-the-de-facto-standard-for-a-reverse-proxy-to-tell-the-backend-ssl-is-used) are in use. And, there's an [RFC to standardize headers](https://tools.ietf.org/html/rfc7239).

So, for the moment, I think it's safe & fine to stick with X-Forwarded-Protocol.

thraxil: FWIW, this is already set up and configured by default in `ccnmtlsettings`.

PR #491, opened 2016-01-19
Fix tinymce error when editing assignment - PMT #104751

nikolas:
This issue came up because there are two potential tinymce editing
areas on this view - one for the response and one for the assignment.
As an instructor, tinymce is initialized for the assignment, as a
student it's initialized for the response, based on the `can_edit`
variable.

In some cases, this if clause in the onTinyMCEInitialize callback
was false when it should be true, due to multiple instances of
ProjectPanelHandler setting self.panel.

    instance.id === self.panel.context.project.id + '-project-content'

Now, I'm not sure yet why there was a conflict - two different instances
should be isolated from each other. But I addressed the issue by
refactoring: only defining the tinymce settings when necessary, taking
this object out of the constructor.

There's also some minor cleanup here:
* Putting the tinymce.show() calls in an if statement prevents the bad
  bug where the screen goes blank. If for some reason there's another
  issue like this, tinymce just won't be initialized.
* Renamed a variable in panel.js that was referring to a single panel,
  not multiple panels.

### DIFF
--- a/media/js/app/panel.js
+++ b/media/js/app/panel.js
@@ -6,17 +6,17 @@
 
 (function() {
     var PanelFactory = function() {
-        this.create = function(el, $parent, type, panels, space_owner) {
+        this.create = function(el, $parent, type, panel, space_owner) {
             // Instantiate the panel's handler
             var handler = null;
             if (type === 'project') {
-                handler = new ProjectPanelHandler(el, $parent, panels,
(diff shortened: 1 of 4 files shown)


sdreher:
Assignment editors now correctly shows, is editable and saves properly. But, inserting selections does not work. Console shows: Uncaught ReferenceError: addMaterialCitation is not defined

PR #513, opened 2016-02-01
Link to FF extension file on addons.mozilla.org

nikolas:
Now that the Firefox extension has been fully approved by Mozilla, we
can have users install this version instead.

> Dear Author,
> Full approval has been granted. Usually, due to your low user count, it
> is not granted. However I see that this tool is very helpful to students
> in many different colleges. I feel the audience would be great.


sdreher:
cool! great feedback.

PR #517, opened 2016-02-05
Validate Selection Start/End Times

sdreher:
Users are occasionally entering video selection times (HH:MM:SS) incorrectly as hours rather than minutes. e.g. 01:01:00 instead of 00:01:01. When this happens, the start and end times are greater than the video duration. The selection times are ignored and the video plays from the beginning.

This code verifies the start & end times are not larger than the duration where possible. Depending on the player, duration is not always immediately available. For example, Vimeo's duration is not recorded until the player begins playing. But, I think this will catch many errors and save some headaches.

### DIFF
--- a/media/css/mediathread.css
+++ b/media/css/mediathread.css
@@ -2870,6 +2870,12 @@ tr.sherd-clipform-editing td {
     text-align: right;
 }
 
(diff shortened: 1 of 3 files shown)


nikolas:
:+1:

PR #544, opened 2016-03-14
Promote/Demote Users

sdreher:
Instructors can promote and demote users to/from instructor status.


mediathread/templates/dashboard/class_roster.html
```diff
@@ -10,31 +10,42 @@
     <link rel="stylesheet" href="{% static 'js/lib/tablesorter/theme.default.custom.css' %}">
 {% endblock %}
 
+{% block uncompressable_css %}
```
nikolas: Should this be uncompressable_js?

sdreher: The underscore/backbone files are usually pre-loaded in the head.

nikolas: Oh okay, i see that uncompressable_js is at the end of the page, while uncompressable_css adds it to `<head>`. I think that's kind of confusing, if we have javascript libraries that need to be in the uncompressable_css block, so might clean this up in the future.

sdreher: definitely.

nikolas:
This is working well for me

PR #547, opened 2016-03-15
Re-organize course picker - PMT #105456

nikolas:
Updates courseaffils to 2.0.8.

I am comfortable pushing this feature without a feature flag, but
if you disagree I can spend some time implementing the feature flag.
Also, I imagine the feature flag would be tricky to implement since
I would need to return a different view from the courseaffils
middleware...

### DIFF
--- a/mediathread/main/views.py
+++ b/mediathread/main/views.py
@@ -4,7 +4,7 @@ import json
 from courseaffils.lib import in_course_or_404, in_course
 from courseaffils.middleware import SESSION_KEY
 from courseaffils.models import Course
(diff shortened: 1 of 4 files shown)


nikolas:
I'm closing this while I implement the changes we discussed:
* Rename "Test Courses" to "Sandboxes"
* Faculty (non-staff users) need to see the sandboxes list
* Add a search filter at the top of the page that filters both the tables (similar to PMT's dashboard). DataTables has this feature built in, so maybe switch to that from tablesorter.

PR #565, opened 2016-03-23
send_template_email utility function

sdreher:
a little mixin to handle templatized emails.

### DIFF
--- a/mediathread/mixins.py
+++ b/mediathread/mixins.py
@@ -2,21 +2,24 @@ import csv
 import json
 
 from courseaffils.lib import in_course_or_404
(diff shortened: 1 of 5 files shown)


mediathread/mixins.py
```diff
@@ -211,3 +214,11 @@ def dispatch(self, *args, **kwargs):
             return HttpResponseForbidden("forbidden")
         self.project = project
         return super(ProjectEditableMixin, self).dispatch(*args, **kwargs)
+
+
+class EmailMixin(object):
```
thraxil: Why a mixin? Why not just a plain function?

PR #588, opened 2016-04-11
Course activation basics - PMT #105457

nikolas:
These features are behind the 'instructor_homepage' waffle flag

### DIFF
--- a/mediathread/main/views.py
+++ b/mediathread/main/views.py
@@ -2,13 +2,15 @@ from datetime import datetime
 import json
 
 import waffle
-from courseaffils.lib import in_course_or_404, in_course
+from courseaffils.lib import (
+    in_course_or_404, in_course, faculty_courses_for_user
+)
 from courseaffils.middleware import SESSION_KEY
 from courseaffils.models import Course
 from courseaffils.views import get_courses_for_user, CourseListView
 from django.conf import settings
 from django.contrib import messages
-from django.contrib.auth.models import User
+from django.contrib.auth.models import User, Group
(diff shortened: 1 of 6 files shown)


mediathread/main/views.py
```diff
@@ -723,6 +733,30 @@ class HomepageView(LoggedInAsFacultyMixin, CourseListView):
 
     def get_context_data(self, **kwargs):
         context = super(HomepageView, self).get_context_data(**kwargs)
-        activatable_courses = []
+        if not waffle.flag_is_active(self.request, 'instructor_homepage'):
+            return context
+
+        groups = self.request.user.groups.all()
```
sdreher: Picking up all groups will also include non-course based groups associated with collaborations. There's some complexity here as well as to what groups are created for the user at login. Maybe we can walk through this flow together?

mediathread/main/views.py
```diff
@@ -72,6 +78,10 @@ def triple_homepage(request):
     if not request.course:
         return HttpResponseRedirect('/accounts/login/')
 
+    if waffle.flag_is_active(request, 'instructor_homepage') and \
```
sdreher: Placing the redirect here appears to make it impossible to login to a course.

mediathread/main/views.py
nikolas: Yeah.. I wanted to talk to you about this, let me know when you have a chance.

mediathread/main/views.py
nikolas: I'll fix this and come up with a test for that.

PR #593, opened 2016-04-15
Remove local admin media files

nikolas:
Since there's no customizations here, we can just use the copies
included in Django.


sdreher:
On staging & production, MeTh is serving assets statically out of the local /media/ directory. (The app hasn't moved to s3 yet.) collectstatic was used to aggregate all the items into that directory. With no other configuration changes, I think this change will break the admin console. (but maybe I'm missing something?)

nikolas:
Thanks for reminding me that MeTh isn't on s3 yet -- that slipped my mind. I don't think that will affect this change, though. When running mediathread with this change locally, for example, requesting the `/media/admin/css/base.css` file correctly returns the version from django's admin assets, even though the `media/admin` directory in the repo no longer exists.

So I think this will be fine on staging, if it's working locally for you.

I guess that also means we would have to keep this in mind when doing future manual runs of `collectstatic`. I've noticed that our S3-enabled apps have this step as part of the rolf deployment process, but mediathread doesn't. Maybe it could?

sdreher:
Right, locally it's cool. But on staging & production, the media directory is configured explicitly: https://github.com/ccnmtl/mediathread/blob/master/apache/sample.conf#L20. I am pretty sure it won't work properly. Let's experiment on QA on Monday. Jonah got kodos back up & running.

thraxil:
The way things are *supposed* to work is that `collectstatic` gets run on every deploy, but is never run in development. Serving static files from S3 forces that workflow. So I'd recommend moving that way.

sdreher:
Right, Mediathead still follows our prior convention where the static files were checked in. We should be moving to an S3/Linode deployment soon.

nikolas:
I'm closing this because this will be addressed when we switch mediathread to the new s3 setup.

PR #602, opened 2016-04-19
Implement course activation form basics

nikolas:
* I've renamed the ActivatableAffil model to Affil, mainly because
  I wanted to put created_at and modified_at fields on this model,
  which required a new table.
* Still TODO: email notification
* The 'student' group I'm getting with a regex (changing fc to st in
  the affil string). I have a question here - should I do this in a
  smarter way? It's possible to do this with
  courseaffils.columbia.WindTemplate, but that would tie the code
  to Columbia-specific logic.


sdreher:
re: the student group. Swapping fc with st is fine.

PR #604, opened 2016-04-20
Homepage Project List - add pagination

sdreher:


nikolas:
This might be a bug.. not sure yet:
2016-04-20-165751_904x459_scrot

To reproduce:
* go to the CCNMTL Social Work Training Site sandbox course
* In the Author dropdown, select "Me"

"Fetching compositions & assignments" displays in the middle of the page, above the modal dialog, and hangs.

sdreher:
This likely is because you are not a member of the class. The UI should error out appropriately though. I'll add some logic around this operation.

nikolas:
Okay that looks good now. I've run into something else which I'm not sure is a bug, it might be because I'm in the CCNMTL Internal course -- I know that course has problems of its own.

2016-04-21-111452_1229x752_scrot

To reproduce:
* As an instructor (logged in as uni), go to CCNMTL Internal
* In Author dropdown, select All Class Members
* click the pagination arrow for the next page

I get a 500 error from the API:

> /api/project/user/njn2118/?page=2&_=1461251638289 Failed to load resource: the server responded with a status of 500 (INTERNAL SERVER ERROR)

sdreher:
I can reproduce the issue, and will fix.

sdreher:
The delete functionality also needed attention. This is ready for re-review.

media/templates/homepage.mustache
```diff
@@ -80,6 +80,30 @@
...
+<ul class="projectlist pagination">
+{{#previous_page}}
+    <li>
+        <a href="#" aria-label="Previous" data-page="{{.}}" class="page">
+            <span class="glyphicon glyphicon-chevron-left" aria-hidden="true"></span>
+        </a>
+    </li>
+{{/previous_page}}
+    <li class="description">
+        <a href="#" aria-label="{{current_page}}/{{num_pages}}">
```
nikolas: Clicking this element doesn't do anything, and causes some weird style issues with hovering. Is there a reason that this is an `<a>` tag? I think it could just be a div.

sdreher: Good catch. I've removed the link & updated the styling.

nikolas: The link as the top of the page is fixed but this one (at the bottom of the page) still has the link.

nikolas:
Looks like there is a z-index conflict with the Create menu:
2016-04-28-143914_495x304_scrot

nikolas:
I pushed a fix for the z-index problem.

sdreher:
Thx. I was just fixing this (again). Appreciate the close review.

PR #657, opened 2016-05-09
Redirect Safari traffic to https://

sdreher:
Safari/YouTube are not working properly together when a video is embedded with the Javascript api in an insecure context.

This clientside redirect is a short-term workaround (10 days or so) until the end of the semester when we will switch all traffic to https:// via the standard nginx mechanisms.

### DIFF
--- a/mediathread/templates/base.html
+++ b/mediathread/templates/base.html
@@ -272,6 +272,11 @@
 
                 /* Select from list: Chrome, Firefox, Safari, MSIE */
                 shieldbrowser({"Chrome": 1, "Firefox": 3.6, "Safari": 4, "MSIE": 10 });
+                if (BrowserDetect.browser === 'Safari' &&
+                        window.location.protocol !== 'https:') {
+                    window.location =
+                        window.location.href.replace('http://', 'https://');
+                }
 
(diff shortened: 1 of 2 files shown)


mediathread/templates/base.html
```diff
@@ -272,6 +272,11 @@ <h4 id="logo_ccnmtl">
 
                 /* Select from list: Chrome, Firefox, Safari, MSIE */
                 shieldbrowser({"Chrome": 1, "Firefox": 3.6, "Safari": 4, "MSIE": 10 });
+                if (BrowserDetect.browser === 'Safari' &&
+                        window.location.protocol !== 'https:') {
+                    window.location =
+                        window.location.href.replace('http://', 'https://');
+                }
```
nikolas: This looks good, I'll merge when tests pass, but I would also like to look into doing this in django middleware instead of javascript.

sdreher:
Yeah, I went back and forth on how to get this done quickly. Clientside seemed the simplest/easiest approach. Easy enough to do in middleware as well by sniffing the user agent from the request headers.

nikolas:
Since this is temporary anyways it's really not a big deal. I agree it's pretty simple client-side so it's fine.

PR #666, opened 2016-05-11
revive the docker/docker-compose setup

thraxil:
I had previously commented out the test running step of building the
docker container because Mediathread's tests (for some undetermined
reason) would consistently fail on my desktop, so it was the only way I
could get it to build. That issue has resolved itself though, so I put
that back in as a nice sanity check within the container.

I've also updated the documentation a bit with more details on how to
use docker-compose to run a development environment.

Lastly, I tend to run into issues with running Mediathread locally
because of the Pickle session serializer, which always breaks because my
browser generally already has a (JSON) session cookie for my local
setup. Clearing out the cookies is a pain, so I set the serializer in
the compose settings to JSON serializer to make it convenient to do
development against it. Since no one else is using the compose setup yet
that I know of, it shouldn't break anyone else's setup.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ RUN /ve/bin/pip install --no-index -f /wheelhouse -r /wheelhouse/requirements.tx
 && rm -rf /wheelhouse
 WORKDIR /app
 COPY . /app/
-#RUN /ve/bin/python manage.py test
+RUN /ve/bin/python manage.py test
 EXPOSE 8000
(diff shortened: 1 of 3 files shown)


thraxil:
:smiling_imp: 666, the PR of the beast.

sdreher:
and a new merge option appears

screen shot 2016-05-11 at 9 20 43 am

mrenoch:
Offering up some :rooster: and :goat:s!

PR #671, opened 2016-05-12
Instructor Dashboard basics - PMT #105469

nikolas:
* Each dashboard view (e.g. settings, sources) now has an absolute url
  containing the course ID
* The "Manage" menu will change to an "Instructor Dashboard" link with
  the instructor_dashboard flag enabled.

The Instructor Dashboard page itself if plain so far - just linking out
to the various functionality. I'll clean this up in future PRs.

### DIFF
--- a/media/css/mediathread.css
+++ b/media/css/mediathread.css
@@ -4362,3 +4362,7 @@ a.btn.btn-default img.left {
     background-color: white;
     color: #337ab7;
 }
+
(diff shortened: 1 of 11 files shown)


nikolas:
So I'm going to have to update the tests and the front-end code to move forward with these changes. But maybe we should talk about this because it might not be the right direction to go in. In any case, I'm closing this PR for now.

PR #677, opened 2016-05-14
Selection Assignment: require due date

sdreher:


mediathread/factories.py
```diff
@@ -326,7 +326,8 @@ def setup_sample_selection_assignment(self):
         assignment = ProjectFactory.create(
             title='Sample Selection Assignment',
             course=self.sample_course, author=self.instructor_one,
-            policy='CourseProtected', project_type='selection-assignment')
+            policy='CourseProtected', project_type='selection-assignment',
+            due_date=datetime.today() + timedelta(1))
```
thraxil: FWIW, I don't think most of us know off the top of our heads that `timedelta`'s default argument is days. I prefer to always just be explicit so it's a little more obvious: `timedelta(days=1)`

PR #685, opened 2016-05-19
Course Dashboard Settings view - PMT #106652

nikolas:


mediathread/urls.py
```diff
@@ -214,10 +214,15 @@
     # Composition Space
     (r'^project/', include('mediathread.projects.urls')),
 
-    # Instructor Dashboard & reporting
+    # Instructor Dashboard
     url(r'^course/(?P<pk>\d+)/dashboard/$',
```
sdreher: The course prefix here is a bit problematic due to the way Courseaffils middleware is activated. Right now, "course" is considered an anonymous path, and as such the middleware does not decorate the request with the course. The base template and LoggedInCourseMixin currently expect request.course to be available.

If you're working on eliminating the decoration (yay!), you'll need to look at assumptions in the base templates and the mixins.

mediathread/main/views.py
```diff
@@ -787,17 +788,40 @@ def get_success_url(self):
...
 
     def get_context_data(self, *args, **kwargs):
-        ctx = super(CourseInstructorDashboardView, self).get_context_data(
+        ctx = super(InstructorDashboardView, self).get_context_data(
             *args, **kwargs)
         ctx.update({'course': ctx.get('object')})
         return ctx
 
 
+class InstructorDashboardSettingsView(
```
sdreher: The view should check that the logged in user is faculty for the requested course.

mediathread/main/views.py
```diff
@@ -787,17 +788,40 @@ def get_success_url(self):
         return reverse('course-invite-complete')
 
 
-class CourseInstructorDashboardView(LoggedInMixin, DetailView):
+class InstructorDashboardView(LoggedInMixin, DetailView):
```
sdreher: The view should check that the logged in user is faculty for the requested course.

mediathread/templates/courseaffils/course_update_form.html
```diff
@@ -0,0 +1,36 @@
+{% extends 'base.html' %}
+{% load bootstrap3 %}
+
+{% block title %}&mdash; Course Settings{% endblock %}
+
+{% block content %}
+    <h1>Course Settings: {{object.title}}</h1>
```
sdreher: This can be an h2 to match the rest of the dashboard pages. {{object.title}} can be removed.

mediathread/templates/main/instructor_dashboard.html
```diff
@@ -11,7 +11,8 @@
```
sdreher: The template incorrectly overrides the coursetitle block. Can we keep the look/feel of this page consistent with the rest of the dashboard pages until Z gets a chance to work her magic?

PR #688, opened 2016-05-19
Fix bug where instructor name is not present on activated course

nikolas:


mediathread/main/views.py
```diff
@@ -891,6 +891,7 @@ def create_course(self, form, affil):
 
         # Add the current user as an instructor.
         c.faculty_group.user_set.add(self.request.user)
+        c.add_detail('instructor', self.request.user.profile.fullname)
```
sdreher: The courseaffils library has a [get_public_name](https://github.com/ccnmtl/django_courseaffils/blob/master/courseaffils/lib.py#L64) function that should be preferred here. (Also available as a template tag [`{% public_name for user %}`](https://github.com/ccnmtl/django_courseaffils/blob/master/courseaffils/templatetags/coursetags.py#L119). The anonymize functionality is handled in that layer.

nikolas: Thanks for pointing that out! I've updated the code.

nikolas: We are displaying instructor names with `User.get_full_name` on the CourseListView: https://github.com/ccnmtl/mediathread/blob/master/mediathread/templates/courseaffils/course_row.html#L23  Should I use that instead of `get_public_name`?

nikolas: Nevermind, that case is only used when the row is an Affil, not a course. Still, it doesn't seem right to store the anonymized name.

PR #749, opened 2016-07-14
Allow add unis per line

thraxil:
both of these recent sentry errors: 

* https://sentry.ccnmtl.columbia.edu/sentry-internal/mediathread/group/1131/
* https://sentry.ccnmtl.columbia.edu/sentry-internal/mediathread/group/1132/

Seem to come from someone entering unis one per line rather than comma separated. This PR adds support for that. The UI still says to use commas, but I think it's OK to have it silently support linebreaks as well since it's pretty clear what the user's intent was (it actually supports any whitespace as a valid separator). We could go the other way and be stricter about requiring commas, but to do that properly, I think it would have to be done with client-side validation and would be a bit more work.


sdreher:
+1. Thanks for resolving this.

PR #809, opened 2016-08-12
Update course activation faculty email

nikolas:


mediathread/main/views.py
```diff
@@ -804,15 +804,23 @@ def send_faculty_email(form, faculty_user):
...
-Consult or Demo: {}
-""".format(data.get('course_name'),
-           data.get('term'),
-           data.get('year'),
-           data.get('consult_or_demo'))
-
+Dear {},
+
+Thank you for creating your Mediathread course {}
+https://mediathread.ccnmtl.columbia.edu/
```
sdreher: Is this url supposed to be here?

mediathread/main/views.py
```diff
@@ -804,15 +804,23 @@ def send_faculty_email(form, faculty_user):
...
+You are now ready to get started. Documentation is online here:
+http://support.ccnmtl.columbia.edu/knowledgebase/topics/6593-mediathread
+
+If you are new to Mediathread, a CTL learning designer or your
+department specialist will check in with you in the coming days to
+make sure all is going well. If you have any pressing questions in the
+meantime, please feel free to contact us at
+mediathread@ccnmtl.columbia.edu.
```
sdreher: I think our address is: ccnmtl-mediathread@ccnmtl.columbia.edu

mediathread/main/views.py
nikolas: I just copied the format that was in the google document. I guess it doesn't really make sense to have it here, I will remove it. I think Mark was implying that we want to include a link to the mediathread course, courses don't have addressable URLs yet.

PR #811, opened 2016-08-12
add top border to course table

nikolas:


media/css/mediathread.css
```diff
@@ -803,6 +803,7 @@ table.course-choices {
     padding:5px;
     margin-left: 10px;
     width: 960px;
+    border-top: 1px solid #ddd
```
sdreher: missing semi-colon

PR #813, opened 2016-08-12
Course activation form style changes

nikolas:


mediathread/main/forms.py
```diff
@@ -198,7 +198,9 @@ def clean(self):
 
 class CourseActivateForm(forms.Form):
     affil = forms.IntegerField(widget=forms.HiddenInput())
-    course_name = forms.CharField(label='Course Name')
+    course_name = forms.CharField(
+        label='Course Name (required)',
```
sdreher: The Course Settings page refers to this field as "Course Title". Can we make them consistent?

PR #815, opened 2016-08-12
Add Create new course button to My Courses page

nikolas:


mediathread/templates/main/course_list.html
```diff
@@ -52,6 +50,11 @@
             <li class="{% if semester_view == 'future' %}active{% endif %}">
                 <a id="future_semesters" href="?semester_view=future">Future</a>
             </li>
+            <li class="navbar-right">
```
sdreher: conditional needed here: staff only

nikolas: Thanks, I added an `add_privilege` check just like the existing course list page uses.

PR #820, opened 2016-08-15
Add course activation message - PMT #107321

nikolas:


nikolas:
I've also added redirection to the correct 'My Courses' tab on course activation on this branch, because it's related.

PR #822, opened 2016-08-16
Add PMT item on course activation - PMT #107330

nikolas:


mediathread/main/views.py
```diff
@@ -811,7 +814,8 @@ def send_faculty_email(form, faculty_user):
             subject,
             body,
             settings.SERVER_EMAIL,
-            [faculty_user.email])
+            [faculty_user.email],
+            fail_silently=not settings.DEBUG)
```
sdreher: The way I'm reading this: fail_silently is set to True in production. Is that the intention?

nikolas: The intention was to not give a 500 error to the user on production. Maybe I should remove this and instead catch the exception.

sdreher: Yeah, I mean, ideally, we'd somehow want to know about the error while still allowing the user to continue with the flow.

PR #828, opened 2016-08-16
Remove unused course_list template.

nikolas:


nikolas:
Alright.. so it looks like the selenium tests are somehow connecting to the `courseaffils/course_list.html` template, but dev & production are using `main/course_list.html`. I'm glad we uncovered this weirdness! I'll investigate.

PR #829, opened 2016-08-16
Fix table display issue for courses with no term

nikolas:
Sandboxes have no term, which was causing a display issue in the
sandboxes tab.


nikolas:
Here's a screenshot of the problem - see how the labels are off:
2016-08-16-114040_997x199_scrot

PR #837, opened 2016-08-17
Catch SSLError when making PMT item

nikolas:


nikolas:
I ran into this exception when testing this locally with my self-signed SSL cert. Locally, it can be worked around by adding `verify=False` to the requests.post() call, but if it ever happens on production we can catch this error, hiding it from the user.

PR #839, opened 2016-08-18
Remove main/course_list.html - use courseaffils/course_list.html.

nikolas:
Selenium tests where using the slightly different
courseaffils/course_list, while dev/prod was using main/course_list.
This change makes both these environments use the same template, and
removes the redundant template.

This confusion all came about because of the course_activation feature
flag, which I'm removing here because it's more trouble than it's worth.


nikolas:
Really, it was my fault that I used the feature flag to replace the entire template instead of just the course activation row. I forgot why I did this - maybe I just wasn't thinking straight.

PR #866, opened 2016-09-12
Remove MochiKit-related stuff from sherdJS

nikolas:


sdreher:
But what if we want to switch back?! :wink:

nikolas:
yeah good point

PR #867, opened 2016-09-12
Add download button to flowplayer view - PMT #107724

nikolas:
This is currently hidden as I'm figuring out how to only display this
for instructors.

2016-09-12-165341_547x487_scrot

### DIFF
--- a/media/css/mediathread.css
+++ b/media/css/mediathread.css
@@ -4390,6 +4390,7 @@ ul.section-tabs>li>a.active {
 }
 
 .sherd-flowplayer-download-btn {
-    display: none;
-    margin-right: 10px;
+    font-size: 1.5em;
+    margin: 10px;
+    float: right;
 }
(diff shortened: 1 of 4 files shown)


sdreher:
Another possible approach: Render the button in the [asset_workspace.html mustache template](https://github.com/ccnmtl/mediathread/blob/master/media/templates/asset_workspace.mustache#L60). User role may already be part of the context.

nikolas:
Hmm good point. I did this in the flowplayer setup code because I need to attach to an event there to get the URL, but I could put the empty button wrapper in the mustache template and still attach to it. I'll see if I can get the role from there.

media/css/mediathread.css
```diff
@@ -4390,6 +4390,7 @@ ul.section-tabs>li>a.active {
 }
 
 .sherd-flowplayer-download-btn {
-    display: none;
-    margin-right: 10px;
+    font-size: 1.5em;
```
sdreher: hide by default for non-flowplayer assets?

PR #902, opened 2016-10-05
JuxtaposeAsset migrations

nikolas:
This adds the model that will be used for the Juxtaposition responses.

### DIFF
--- a/mediathread/juxtapose/models.py
+++ b/mediathread/juxtapose/models.py
@@ -1,0 +1,35 @@
+from django.db import models
+from django.contrib.auth.models import User
+from courseaffils.models import Course
+from mediathread.djangosherd.models import SherdNote
+
+
+class JuxtaposeAsset(models.Model):
+    added = models.DateTimeField(auto_now_add=True)
+    modified = models.DateTimeField(auto_now=True)
+    author = models.ForeignKey(User)
+    course = models.ForeignKey(Course)
+    title = models.TextField()
+    spine = models.ForeignKey(SherdNote, blank=True, null=True)
+
+
+class JuxtaposeMediaElement(models.Model):
+    added = models.DateTimeField(auto_now_add=True)
+    modified = models.DateTimeField(auto_now=True)
+    juxtaposition = models.ForeignKey(JuxtaposeAsset)
+    start_time = models.DecimalField(
+        max_digits=8, decimal_places=5)
+    end_time = models.DecimalField(
+        max_digits=8, decimal_places=5)
+    media = models.ForeignKey(SherdNote)
+
+
+class JuxtaposeTextElement(models.Model):
+    added = models.DateTimeField(auto_now_add=True)
+    modified = models.DateTimeField(auto_now=True)
+    juxtaposition = models.ForeignKey(JuxtaposeAsset)
+    start_time = models.DecimalField(
+        max_digits=8, decimal_places=5)
+    end_time = models.DecimalField(
+        max_digits=8, decimal_places=5)
+    text = models.TextField()
--- a/mediathread/assetmgr/tests/test_models.py
+++ b/mediathread/assetmgr/tests/test_models.py
@@ -4,10 +4,12 @@ from django.test import TestCase
 
 from mediathread.assetmgr.models import Asset, Source
 from mediathread.djangosherd.models import SherdNote
-from mediathread.factories import MediathreadTestMixin, AssetFactory, \
-    UserFactory, SherdNoteFactory, ProjectFactory, \
-    SuggestedExternalCollectionFactory, SourceFactory, \
-    ExternalCollectionFactory
+from mediathread.factories import (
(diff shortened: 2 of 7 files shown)


mediathread/assetmgr/models.py
```diff
@@ -339,3 +339,22 @@ def __unicode__(self):
...
     class Meta:
         ordering = ['title']
+
+
+class JuxtaposeAsset(models.Model):
+    added = models.DateTimeField(auto_now_add=True)
+    modified = models.DateTimeField(auto_now=True)
+    author = models.ForeignKey(User)
+    course = models.ForeignKey(Course)
+    title = models.TextField(blank=True, null=True)
```
sdreher: ideally, not blank or null

mediathread/assetmgr/models.py
```diff
@@ -339,3 +339,22 @@ def __unicode__(self):
...
+    spine = models.ForeignKey(Source, blank=True, null=True)
```
sdreher: linking to a Sherdnote here would fit better with the current architecture, and make this more akin to a straight Asset.

mediathread/assetmgr/models.py
```diff
@@ -339,3 +339,22 @@ def __unicode__(self):
...
+
+
+class JuxtaposeItem(models.Model):
```
sdreher: Unclear on how this object fits in?

nikolas: This object is intended to represent an item on one of the Juxtapose tool's tracks. I've split this out into `JuxtaposeMediaItem` and `JuxtaposeTextItem`, since they are different enough that I think that makes sense. I'm planning on renaming the 'aux' track to the 'media' track in Juxtapose.

mediathread/assetmgr/models.py
nikolas: I was thinking of the title as equivalent to the Juxtaposition's "global annotation", which is why I had it optional. I will make this required.

mediathread/assetmgr/models.py
```diff
@@ -339,3 +340,30 @@ def __unicode__(self):
...
+    spine = models.ForeignKey(SherdNote, blank=True, null=True)
+
+
+class JuxtaposeMediaItem(models.Model):
+    added = models.DateTimeField(auto_now_add=True)
+    modified = models.DateTimeField(auto_now=True)
+    juxtaposition = models.ForeignKey(JuxtaposeAsset)
+    range1 = models.FloatField()
+    range2 = models.FloatField()
+    source = models.ForeignKey(Source)
```
sdreher: SherdNote FK here as well.

mediathread/assetmgr/models.py
```diff
@@ -339,3 +340,30 @@ def __unicode__(self):
...
+class JuxtaposeAsset(models.Model):
+    added = models.DateTimeField(auto_now_add=True)
+    modified = models.DateTimeField(auto_now=True)
+    author = models.ForeignKey(User)
+    course = models.ForeignKey(Course)
+    title = models.TextField()
+    spine = models.ForeignKey(SherdNote, blank=True, null=True)
+
+
+class JuxtaposeMediaItem(models.Model):
```
sdreher: Putting on my Mark hat here: I'm concerned by overloading the terms "Item", "range1" and "range2".

"Item" refers to something very specific in Mediathread, e.g. a whole video, image or audio track. Perhaps "Clip" instead?

"range1" and "range2" is how the code denotes the portion of the larger item or asset. Perhaps "cue1" and "cue2"?

mediathread/assetmgr/models.py
nikolas: Alright, I understand now, looking at the diagram, and I will make that change. Currently there's a circular dependency problem because the file that SherdNote is defined in imports Asset, so I might have to put these models somewhere else.

mediathread/juxtapose/models.py
```diff
@@ -0,0 +1,31 @@
...
+    course = models.ForeignKey(Course)
+    title = models.TextField()
+    spine = models.ForeignKey(SherdNote, blank=True, null=True)
+
+
+class JuxtaposeMediaItem(models.Model):
+    added = models.DateTimeField(auto_now_add=True)
+    modified = models.DateTimeField(auto_now=True)
+    juxtaposition = models.ForeignKey(JuxtaposeAsset)
+    start_time = models.FloatField()
```
nikolas: I'm wondering if the inexact nature of the float type is going to cause weirdness when dealing with items that occur directly after each other, with no space in between. I will switch this to Django's [DecimalField](https://docs.djangoproject.com/en/1.10/ref/models/fields/#decimalfield) so we have the best control over these timestamps.

mediathread/assetmgr/models.py
nikolas: Also, I will rename "Item" to something else. I appreciate the detailed feedback.

PR #917, opened 2016-10-19
Add djangorestframework API for juxtapose models

nikolas:
I'm leaving the API just read-only for now as I connect it
to the javascript application.


sdreher:
Reminder: Item & selection content must be filtered through authentication and authorization layers.

PR #926, opened 2016-10-25
Add hook for juxtapose asset creation / update

nikolas:


mediathread/projects/models.py
```diff
@@ -628,4 +628,13 @@ class ProjectNote(models.Model):
 class ProjectJuxtaposeAsset(models.Model):
     """This model connects the JuxtaposeAsset to the Project."""
     juxtapose_asset = models.ForeignKey(JuxtaposeAsset)
+    # Points to an "Assignment" Project.
     project = models.ForeignKey(Project)
+
+    @classmethod
+    def get_asset_for_user(cls, project, user):
```
thraxil: I think this would be better as a method on a Manager, rather than a classmethod directly on the model.

nikolas: That thought crossed my mind, but I couldn't think of any real benefits to having it in the model manager. Maybe it just makes more sense to be there, though. I'll make that change.

nikolas: It turns out this function wasn't even necessary. I've simplified it to a call to `ProjectJuxtaposeAsset.objects.filter()`

PR #936, opened 2016-10-28
Add media_type filter to /api/asset/

nikolas:
This adds a media_type query param to the `/api/asset/` endpoint. It's
implement in an ad-hoc way, outside of [tastypie's built-in filtering system](http://django-tastypie.readthedocs.io/en/latest/resources.html#basic-filtering),
because the way I've implemented it is simple and works for our use
case.

I've also removed the AjaxRequiredMixin from the view, which we can talk
about if you want. I don't believe requiring AJAX is necessary for an
API, and I find it nice to be able to explore the API by just entering
the URL in the browser.


nikolas:
I've made the changes we discussed on Friday afternoon - `filter_by_media_type()` is now in djangosherd/models.py along with the other note filtering code.

PR #944, opened 2016-11-01
Add view_callback to CollectionWidget

nikolas:
This allows me to register the click events for adding selections to the
sequence assignment.


sdreher:
Would listening events be a better, less tightly bound approach?

nikolas:
Yeah, I was just thinking over how I would bind to the dom elements from the callback, and it would be difficult to get the data back to React. I will look into broadcasting some event from the CollectionWidget that I can subscribe to within react.

PR #953, opened 2016-11-04
Add some basic tests for the sequence assignment API

nikolas:


mediathread/sequence/tests/test_apiviews.py
```diff
@@ -0,0 +1,77 @@
...
+    def test_list(self):
+        SequenceAssetFactory()
+        SequenceAssetFactory()
+
+        r = self.client.get(reverse('sequenceasset-list'))
+        self.assertEqual(r.status_code, 403)
+
+    def test_retrieve(self):
+        return
+        # asset = SequenceAssetFactory()
```
thraxil: Any particular reason for keeping the commented tests in here?

PR #1001, opened 2016-11-29
Sequence api: allow anonymous access

nikolas:


mediathread/sequence/apiviews.py
```diff
@@ -27,12 +26,12 @@ def perform_create(self, serializer):
 
 
 class SequenceMediaElementViewSet(viewsets.ModelViewSet):
-    permission_classes = (permissions.IsAuthenticated,)
+    permission_classes = (ReadOnly,)
```
sdreher: The permission here should be based on whether the user has chosen to publish the parent project to the world. Otherwise, the isAuthenticated gate remains. Likely you'll need to add some custom authentication here.

PR #1054, opened 2016-12-20
Add events for save and setDirty - PMT #109033, #109142

nikolas:


media/js/app/projects/sequenceAssignmentView.js
```diff
@@ -61,6 +62,19 @@
...
                 jQuery(window).bind('beforeunload', this.beforeUnload);
             }
+
+            this.mapSignals();
+        },
+        mapSignals: function() {
+            var self = this;
+
+            jQuery(window).on('sequenceassignment.save', function(e) {
+                self.onSaveProject();
```
nikolas: I was envisioning it the other way around, but the way you've described makes sense. I will change this from a listener to a trigger and call it in `onSaveProject.success`.

sdreher: Cool. Let's go over it offline if it doesn't make sense as you code it up.

media/js/app/projects/sequenceAssignmentView.js
```diff
@@ -148,6 +159,9 @@
                     $saveButton.removeAttr('disabled')
                         .removeClass('saving', 1200, function() {
                             jQuery(self).text('Saved'); });
+
+                    document.dispatchEvent(
```
nikolas: makes sense - working on that now.

sdreher:
Excellent. I wasn't able to verify a success route due to some component issue, but the trigger to React and back on the error route is working properly. This looks good.

PR #1070, opened 2016-12-22
Only allow sequence submission when valid - PMT #109151

nikolas:
This prevents the Submit button from being active when the sequence is
in a non-submittable state.

We'll also be setting the submittable state from React as the project
loads, during initialization. This is for the case where the response
is in a submittable state, but the user has just navigated to it without
doing any editing or saving.

### DIFF
--- a/media/js/app/projects/sequenceAssignmentView.js
+++ b/media/js/app/projects/sequenceAssignmentView.js
@@ -78,11 +78,18 @@
                     self.setDirty(data.dirty);
                 });
 
+            jQuery(window).on(
+                'sequenceassignment.set_submittable',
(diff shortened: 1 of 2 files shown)


media/js/app/projects/sequenceAssignmentView.js
```diff
@@ -123,18 +123,22 @@
         isDirty: function() {
             return this.dirty;
         },
-        setDirty: function(isDirty) {
+        setDirty: function(isDirty, isSubmittable) {
```
nikolas: On second thought, I'm going to try breaking this out into two methods: `setDirty` and `setSubmittable`.

PR #1101, opened 2016-12-28
CollectionWidget: simplify save signal

sdreher:
Just send over assetId & annotationId on save. 

This PR is related to this one: https://github.com/ccnmtl/juxtapose/pull/195

### DIFF
--- a/media/js/app/assetmgr/collectionwidget.js
+++ b/media/js/app/assetmgr/collectionwidget.js
@@ -256,9 +256,9 @@ CollectionWidget.prototype.mapEvents = function() {
                 success: function(json, textStatus, xhr) {
                     self.signalInsert(assetId, json.annotation.id);
                     self.$modal.modal('hide');
-                    return;
                 }
             });
+            return;
         }
 
         // get or create the user's global annotation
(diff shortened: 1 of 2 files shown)


media/js/app/assetmgr/collectionwidget.js
```diff
@@ -256,9 +256,9 @@ CollectionWidget.prototype.mapEvents = function() {
                 success: function(json, textStatus, xhr) {
                     self.signalInsert(assetId, json.annotation.id);
                     self.$modal.modal('hide');
-                    return;
                 }
             });
+            return;
```
nikolas: This looks fine really, but I was under the assumption that an empty `return` at the end of an event handler is unnecessary. In my mind both of these should be removed - I'm aware of returning `false` from an event handler as discussed here: http://stackoverflow.com/a/4379459/173630 but what does this `return` do?

sdreher: The return at Ln. 261 is needed to prevent the code from falling into the next bit of logic and creating a global annotation when one already exists. Noting though - the return at Ln 276 is unneeded and can be removed. I'll do a PR to fix that. Thanks for the close review.

nikolas: Ah okay, makes sense - this return isn't actually at the end of the event handler.

PR #1117, opened 2017-01-03
Use normal range input for playhead - PMT #109411

nikolas:
This resolves the issue in the referenced PMT, and makes the playhead easier to use. I made the play button not extend all the way down to the tracks, which I admit looks a little odd, but the reason is that different browsers will have slightly different heights for their range inputs, so we can't use an absolute value for the play button height if it needs to extend to the tracks. 

Mac / Chrome
screen shot 2017-01-03 at 14 49 08

Mac / Firefox:
screen shot 2017-01-03 at 14 49 20

Depends on changes from this juxtapose branch: https://github.com/ccnmtl/juxtapose/pull/214

### DIFF
--- a/media/css/mediathread.css
+++ b/media/css/mediathread.css
@@ -2948,10 +2948,6 @@ textarea {
     border: 1px solid #ccc;
 }
 
-input {
-    border: 1px solid #333333;
-}
-
 input[type="checkbox"] {
     border: none; /* IE 8,9 look very very odd with a border */
 }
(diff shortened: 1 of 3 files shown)


sdreher:
The functionality on this is considerably better than the old. One suggestion - would you be opposed to hiding that range indicator, like so? (We also still need to address that number overhang.)

screen shot 2017-01-04 at 9 53 22 am

nikolas:
Yeah, that could be a good way to clean this up, as long as the functionality remains intact where you can click on a time point to seek to it.

sdreher:
Okay, I'll do that temporarily on this branch. Then we can get @zmustapha in to offer some more suggestions to polish it.

PR #1119, opened 2017-01-04
playhead: minor style tweaks

sdreher:
screen shot 2017-01-04 at 10 16 07 am

* hide the range. the track can still be clicked to jump to a particular point.
* add some borders

### DIFF
--- a/media/juxtapose/css/playhead.css
+++ b/media/juxtapose/css/playhead.css
@@ -4,6 +4,9 @@
 .jux-playhead-container input[type=range] {
     cursor: col-resize;
     width: 930px;
+    -webkit-appearance: none;
+    background: transparent;
+    outline: none;
 }
 
 .jux-playhead-line {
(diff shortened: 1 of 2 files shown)


media/juxtapose/css/playhead.css
```diff
@@ -4,6 +4,9 @@
 .jux-playhead-container input[type=range] {
     cursor: col-resize;
     width: 930px;
+    -webkit-appearance: none;
```
nikolas: Just a note - this doesn't seem to be working in Firefox, even with `-moz-appearance: none`

sdreher: drat. I was basing it off this: https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/.

nikolas: Oh maybe there's a way that involves something trickier used in that post - I'll look into it.

PR #1128, opened 2017-01-04
Add css transition animation to playhead line

nikolas:


nikolas:
I'm going to hold off on this until I come up with a solution that includes the range input as well: https://pmt.ccnmtl.columbia.edu/item/109439/#event-2

PR #1133, opened 2017-01-05
Tighten up timeline ruler display

nikolas:
This puts the labels below the ruler and saves some vertical space.

Maybe we should hide the `00:00:00` label, since that position is kind
of obvious, and the playhead is covering it in its default state.

screen shot 2017-01-05 at 10 57 07 am


sdreher:
Definitely something to review with Zarina.

PR #1157, opened 2017-01-10
Fix exception when invalid asset is collected - PMT #108503

nikolas:
Show an error instead of raising an exception, in this case.


nikolas:
Actually - this sentry error data can be useful in helping us catch sites that aren't collecting assets correctly with the extension. We should still show a user-facing error in this case instead of a 500, but I'll add something so we still get this exception.

PR #1168, opened 2017-01-12
Filter out Vimeo assets from CollectionWidget - PMT #109584

nikolas:
I made the primary_type filter accept an array as well as a single
value. It looks like we need to filter out 'image_fpxid' (artstor) as well
as the 'vimeo' primary type.

This is my first foray into the asset filtering code and the
RestrictedMaterialsMixin so please review and edit or make suggestions as
needed.

### DIFF
--- a/mediathread/mixins.py
+++ b/mediathread/mixins.py
@@ -106,8 +106,14 @@ class RestrictedMaterialsMixin(object):
             self.all_items_are_visible, tag_string, modified, vocabulary)
         visible_notes = visible_notes.filter_by_media_type(
             request.GET.get('media_type'))
-        visible_notes = visible_notes.exclude_primary_type(
-            request.GET.get('primary_type'))
+
+        if request.GET.get('primary_type'):
+            visible_notes = visible_notes.exclude_primary_type(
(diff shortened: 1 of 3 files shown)


mediathread/mixins.py
```diff
@@ -106,8 +106,14 @@ def visible_assets_and_notes(self, request, assets):
...
             request.GET.get('media_type'))
-        visible_notes = visible_notes.exclude_primary_type(
-            request.GET.get('primary_type'))
+
+        if request.GET.get('primary_type'):
+            visible_notes = visible_notes.exclude_primary_type(
+                request.GET.get('primary_type'))
+        elif request.GET.getlist('primary_type[]'):
+            filtered_types = request.GET.getlist('primary_type[]')
+            for t in filtered_types:
```
sdreher: This section could be simplified by modifying the SherdNote.exclude_primary_type function to accept an array always. The query just needs to be updated at [Ln 130](https://github.com/ccnmtl/mediathread/blob/master/mediathread/djangosherd/models.py#L130)  like so: 
```return self.exclude(asset__source__label__in=primary_type)```

PR #1244, opened 2017-02-21
Add standalone sequence composition - PMT #109928

nikolas:
The create link is currently only available to superusers.

Requires: https://github.com/ccnmtl/juxtapose/pull/281

### DIFF
--- a/mediathread/projects/views.py
+++ b/mediathread/projects/views.py
@@ -396,27 +396,14 @@ class SequenceEditView(LoggedInCourseMixin, ProjectReadableMixin,
     template_name = 'projects/sequence.html'
 
     def get(self, request, *args, **kwargs):
-        """
-        A multi-panel editable view for the specified project
-        Legacy note: Ideally, this function would be named project_view but
-        StructuredCollaboration requires the view name
-        to be  <class>-view to do a reverse lookup
-
-        Panel 1: Parent Assignment (if applicable)
-        Panel 2: Project
-        Panel 3: Instructor Feedback (if applicable & exists)
-
-        Keyword arguments:
-        project_id -- the model id
-        """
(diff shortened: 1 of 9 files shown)


mediathread/projects/views.py
```diff
@@ -390,6 +390,36 @@ class SequenceAssignmentEditView(AssignmentEditView):
...
 
 
+class SequenceEditView(LoggedInCourseMixin, ProjectReadableMixin,
+                       TemplateView):
+
+    template_name = 'projects/sequence.html'
+
+    def get(self, request, *args, **kwargs):
+        """
+        A multi-panel editable view for the specified project
```
sdreher: These legacy comments can be removed.

mediathread/projects/views.py
```diff
@@ -390,6 +390,36 @@ class SequenceAssignmentEditView(AssignmentEditView):
...
+
+        Panel 1: Parent Assignment (if applicable)
+        Panel 2: Project
+        Panel 3: Instructor Feedback (if applicable & exists)
+
+        Keyword arguments:
+        project_id -- the model id
+        """
+        project = get_object_or_404(Project, pk=kwargs.get('project_id', None))
+        show_feedback = kwargs.get('feedback', None) == "feedback"
```
sdreher: Feedback and the notion of a "response" are applicable only to the assignment structure. These can be removed here.

sdreher:
This looks good overall. The only missing piece I see is the ability to save at various levels: private, public, public to world.

nikolas:
Right, yeah I remember we talked about that step. I will add that next. We want it to only prompt the user for the visibility level the first time they save, right?

sdreher:
No - levels should be available at all times. Akin to composition save.

PR #1389, opened 2017-06-06
:arrow_up: django.mk 1.4.0

thraxil:


nikolas:
So we are using a custom package for lettuce in Mediathread: https://github.com/ccnmtl/mediathread/blob/master/requirements.txt#L25

I'm all for just using the official pypi for our django apps, we'll just have to keep this in mind for Mediathread, and eventually get off this custom ccnmtl version. I think lettuce might have a new maintainer that is a little more responsive to pull requests.

PR #1441, opened 2017-07-28
Update README markdown

sdreher:


nikolas:
what is this changing? this PR seems like a mistake

nikolas:
Oh nevermind, it looks like you're fixing the layout so these are on different lines now.

PR #1463, opened 2017-08-24
Eslint security checks

sdreher:
* additional eslint security plugins & rules
* cleanup or ignore issues


sdreher:
The selenium test suite is consistently failing on a given test. I will investigate the tweaks I made.

PR #1485, opened 2017-09-08
LTI: determine user role in course

sdreher:
Simplifying the approach.

### DIFF
--- a/lti_auth/templates/lti_auth/fail_course_configuration.html
+++ b/lti_auth/templates/lti_auth/fail_course_configuration.html
@@ -14,7 +14,7 @@
                     <small>
                         Please contact your help desk to enable the component.<br />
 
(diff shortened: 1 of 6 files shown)


lti_auth/templates/lti_auth/fail_course_configuration.html
```diff
@@ -14,7 +14,7 @@
                     <small>
                         Please contact your help desk to enable the component.<br />
 
-                        {% if 'administrator' in roles or 'instructor' in roles %}
+                        {% if is_instructor or is_administrator %}
                             <h3>Administrators & Instructors</h3>
```
nbuonin: @sdreher do you need an html entity here?

sdreher: good catch. yes I do.

PR #1575, opened 2017-11-27
Update broken blakearchive.org links - PMT #112484

nikolas:
This contains updates for two url formats: The direct image URLs and the image
detail URLs.


nbuonin:
I'm not sure if I'm qualified to merge this, but I can understand it and it looks good to me.

PR #1657, opened 2018-02-21
Fix digitaltibet URLs - PMT #113800

nikolas:
See the attached PMT. It slipped my mind that some Source objects just
use their URL to point to the item's permalink. And I've also fixed a
problem where the OpenLayers main image was broken for some assets.

Unfortunately, I'm still seeing some weirdness in OpenLayers. I'm posting this screenshot here in case you have any ideas - I'll continue to investigate. OpenLayers is displaying only the top-left quadrant of this image, when the URL is set to the full image: https://s3.amazonaws.com/ccnmtl-digitaltibet-static-prod/objects/Rubin_C2003_52_1H065108.jpg

2018-02-21-172206_1251x1081_scrot

This is asset 23334 in the "V2365 section 001..." course.


mediathread/assetmgr/migrations/0010_auto_20180221_1700.py
```diff
@@ -0,0 +1,50 @@
...
+        if splitext(filename)[1] == '':
+            # This is a broken permalink. Fix it.
+            permalink = 'https://digitaltibet.ctl.columbia.edu/' + \
+                        'object/{}/'.format(filename.replace('_', '-'))
+            source.url = permalink
+            source.save()
+
+
+def overwrite_with_primary_sources(Source, sources):
+    for source in sources:
```
sdreher: Delving into this issue on my local dev env -- the sources passed through are of type `xyztile`. The end of the source here `x${x}` is a format string that is replaced by a given tile in `sherdjs/src/image/views/openlayers.js`. Replacing these with the primary source url will break the tiling mechanism. I think we'll need to make sure all the tiles are still available in the hugo-fied Digital Tibet, and then just replace the s3 bucket url with the https://digitaltibet.ctl.columbia.edu/ per the `fix_permalinks` function.

nikolas:
I've resolved this by making a change to sherdjs's openlayers viewer instead of making the tiles available on s3. I really didn't want to dig into the drupal 5 instance here, and I don't think they should be using tiles anyways - that's only useful for infinitely scrolling things like maps.

The diff is messy because I cleaned up a bunch of whitespace, but the main change is here in sherdjs/src/image/views/openlayers.js:
```
+            if (create_obj.object.xyztile &&
+                // Always use an untiled version for digitaltibet. This
+                // came about during the migration of this site from
+                // Drupal 5 to Hugo.
+                !/amazonaws.com\/ccnmtl-digitaltibet-static-prod/.test(
+                    create_obj.object.image)
+               ) {
```

With this change, I'm seeing correct annotations on assets like `/asset/40959/annotations/223878/` (in INTRO EAST ASIAN CIV: TIBET V2365- 001 -Fall 2016). On the asset screenshotted above with this change, the "eyebrows" annotation is slightly off. Still - I think this should be merged, as currently on production this asset isn't even showing. Besides, this is an old class - new annotations work fine.

With this sherdjs change:
2018-06-01-143925_552x513_scrot

mediathread-prod:
2018-06-01-143933_597x620_scrot

sdreher:
A quick glance showed some wonky display. I need to spend some time with this.

PR #1710, opened 2018-04-10
Add Dockerfile instructions for smooth execution of 'make build'.

aleksas:
Add Dockerfile instruction to install python-urllib3.
Add Dockerfile instruction to add requirements/ directory otherwise requirements/src/lettuce-0.2.25.ccnmtl-py2.py3-none-any.whl can not be found during make install.


sdreher:
@aleksas - Thanks for your contribution!

PR #1718, opened 2018-04-16
Hide uploader functionality

sdreher:
The upload functionality is only supported at ColumbiaU. Hide the configuration in the base template. Our deploy_specific add the uploader flow.


sdreher:
I need to think about this...

PR #1746, opened 2018-05-21
anonymize GA

nikolas:
Mediathread uses a different GA method - ga.js instead of analytics.js.
I got the anonymizeIp snippet from here:

https://developers.google.com/analytics/devguides/collection/gajs/methods/gaJSApi_gat#_gat._anonymizeIp


sdreher:
So, we're at 4 generations of Google Analytics? We should get an award or something.

PR #1893, opened 2018-11-28
Add some Asset model method documentation

nikolas:
When debugging the Mediathread extension I found some of this stuff
confusing.


mediathread/assetmgr/models.py
```diff
@@ -138,8 +144,7 @@ class Asset(models.Model):
 
     # labels which determine the saving of an asset
     # in order of priority for which label is marked primary
-    # an asset must have at least one source label from this list
-    # 'url' should probably stay at the end
```
nikolas: 'url' is not at the end of `primary_labels`, so I think this comment is out of date

sdreher:
This passed TravisCI, not sure why it still appears yellow. I'm merging.

PR #2133, opened 2019-09-10
Use python 3 by default - ITINF-1608

nikolas:
I've had to customize django.mk a little bit here to still allow for
virtualenv/python2 support for lettuce testing on Travis.


nikolas:
Regarding the zeep 3 upgrade, see: https://github.com/ccnmtl/pypanopto/pull/235

PR #2149, opened 2019-09-18
Use course-aware URL in Asset api - MEDT-3393

nikolas:


mediathread/assetmgr/api.py
```diff
@@ -45,7 +47,17 @@ def apply_filters(self, request, applicable_filters):
     def dehydrate(self, bundle):
         bundle.data['thumb_url'] = bundle.obj.thumb_url
         bundle.data['primary_type'] = bundle.obj.primary.label
-        bundle.data['local_url'] = bundle.obj.get_absolute_url()
+
+        if waffle.flag_is_active(self.request, 'addressable_courses') and \
```
sdreher: Will `waffle.flag_is_active` fail if self.request is None? Maybe check for self.request first?

PR #2251, opened 2019-12-03
Add new course page - MEDT-3450

nikolas:
Add basic tabbed layout for the new course page. Integrate
React with Mediathread, more closely than juxtapose.

Steps to build:
* `npm install`
* `npm run build` or `npm run dev`

I will add jest and cypress for testing, eventually.

This is behind the "new_course_view" feature flag, and can be found at a
separate route from the current course page, at `/react/`, e.g.:
`/course/5/react/`.

2019-12-03-165734_997x888_scrot


media/js/src/main.jsx
```diff
@@ -0,0 +1,124 @@
...
+                            </p>
+                        </div>
+                    </div>
+                );
+            });
+
+            if (assets.length === 0) {
+                assetsDom = 'No assets found.'
+            } else {
+                assetsDom = <ul>{assets}</ul>;
```
nbuonin: This `ul` is wrapping an array of divs, which wrap each asset. Should each asset be an `li`?

media/js/src/main.jsx
```diff
@@ -0,0 +1,124 @@
...
+        let assetsDom = 'Loading Assets...';
+
+        if (this.state.assets && typeof this.state.assets === 'string') {
+            // Display error to user
+            assetsDom = <strong>{this.state.assets}</strong>;
+        } else if (this.state.assets) {
+            this.state.assets.forEach(function(asset) {
+                const thumbnail = asset.thumb_url || asset.sources.image.url;
+                assets.push(
+                    <div className="card" key={asset.id}>
```
nbuonin: I think this might make sense to break this out into its own stateless component.

media/js/src/main.jsx
```diff
@@ -0,0 +1,124 @@
...
+            if (assets.length === 0) {
+                assetsDom = 'No assets found.'
+            } else {
+                assetsDom = <ul>{assets}</ul>;
+            }
+        }
+
+        let tabDom = '';
+        if (this.state.activeTab === 'collection') {
+            tabDom = <div>
```
nbuonin: And maybe break this guy out into its own component too. That way you can conditionally render it in the return. I'm being nit picky, feel free to ignore.

nbuonin:
Looks good @nikolas; I'm going to steal this for Locus Tempus.

One thing that comes to mind is standardizing make files for the js build step. For OHCOE, I have a bunch of make targets for webkit, scss, Cypress, ect; and it all feels a bit messy. I think it might be good if we standardize some of these targets in our Makefiles repo. That's not specific to this PR though, but its something to keep in mind as we start to use React more.

media/js/src/main.jsx
nbuonin: Hell, I might move this if/else into its own component too, along with the `forEach` above to make the logic a little more clear. Again, nit picky, feel free to ignore this too.

nikolas:
Cool, thanks for all the feedback! Yeah, I haven't integrated in the makefile stuff yet, since I'm fine just using npm for now. I guess we would just add targets to `npm run build` and `npm run dev` in `js.mk`.

PR #2279, opened 2020-01-07
Add jest testing setup

nikolas:
Includes basic tests for rendering GridAsset and CollectionTab
with mock data.

Based on the example here:

  https://reactjs.org/docs/testing-recipes.html#data-fetching

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,5 @@ matrix:
 before_install:
   - pip install --upgrade setuptools pip wheel
 script:
+  - make jstest
   - make $TEST_SUITE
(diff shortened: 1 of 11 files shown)


nbuonin:
@nikolas this PR is a helpful example for me. I never really could get my Jest tests to work and I think its because I wasn't using `act`.

nikolas:
Yeah... jest can be tricky to set up since it's flexible and there are many ways to use it, and it depends a lot on your setup. I spent some time stuck setting this up here while configuring babel which files to transform and which ones not to.

This is my first time using the `act` thing, and that's just copy and pasted from React's docs.

PR #2281, opened 2020-01-07
add Asset model with better data error handling

nikolas:


media/js/src/Asset.js
```diff
@@ -0,0 +1,35 @@
+export default class Asset {
+    constructor(data) {
+        this.asset = data;
+    }
+    getType() {
+        let assetType = this.asset.primary_type;
+        if (assetType === 'youtube' || assetType === 'mp4_pseudo') {
```
sdreher: Just a reminder that all the types are [listed here](https://github.com/ccnmtl/mediathread/blob/master/mediathread/assetmgr/models.py#L155). Not sure you're ready for all the things yet!

nikolas: great, thanks! yeah, i will make a note of this in a comment in the front-end code, since these will need to stay in sync.

PR #2290, opened 2020-01-09
Use largest thumbnail available for youtube videos

nikolas:


media/js/src/Asset.js
```diff
@@ -22,6 +24,17 @@ export default class Asset {
...
         }
 
+        // Use youtube's thumbnail format for getting the biggest
+        // thumbnail available.
+        if (
+            this.asset.primary_type === 'youtube' &&
+                this.asset.sources &&
+                this.asset.sources.url
+        ) {
+            const youtubeId = getYouTubeID(this.asset.sources.url.url);
```
sdreher: Is this double `url.url` correct?

nikolas: Yeah, the asset `sources` array has a `url` object:

out

On second glance, maybe I should use youtube.url here? In this case, they both contain the youtube ID so it doesn't matter.

PR #2301, opened 2020-01-14
Search filters: populate tags, terms, owners

nikolas:
* Make title search work with "Go" button, and Enter key, instead
of immediately searching as the user types.

### DIFF
--- a/media/js/src/AssetFilter.jsx
+++ b/media/js/src/AssetFilter.jsx
@@ -6,16 +6,83 @@ export default class AssetFilter extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
-            titleFilter: '',
             selectedOwner: 'all'
         };
-        this.handleChange = this.handleChange.bind(this);
+
+        this.handleOwnerChange = this.handleOwnerChange.bind(this);
+        this.handleTitleSearch = this.handleTitleSearch.bind(this);
     }
-    handleChange(e) {
-        console.log('handleChange');
+    handleTitleSearch(e) {
+        if (e.key === 'Enter') {
+            this.props.handleTitleFilterSearch();
+        }
+    }
+    handleOwnerChange(e) {
+        this.setState({selectedOwner: e.value});
     }
     render() {
         const { selectedOwner } = this.state;
+
+        let ownersOptions = [];
+        if (this.props.assets) {
+            this.props.assets.forEach(function(asset) {
+                const usernames = ownersOptions.map(x => x.value);
+                if (!usernames.includes(asset.author.username)) {
+                    ownersOptions.push({
+                        label: asset.author.public_name,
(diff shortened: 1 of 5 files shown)


media/js/src/AssetFilter.jsx
```diff
@@ -6,17 +6,80 @@ export default class AssetFilter extends React.Component {
...
+            this.props.handleTitleFilterSearch();
+        }
+    }
+    handleOwnerChange(e) {
+        this.setState({selectedOwner: e.value});
     }
     render() {
         const { selectedOwner } = this.state;
 
+        let ownersOptions = [];
```
nbuonin: Because these are working with only props, it might make sense to move this to the constructor so it doesn't run with each call to render.

media/js/src/AssetFilter.jsx
```diff
@@ -6,17 +6,80 @@ export default class AssetFilter extends React.Component {
...
+        if (this.props.assets) {
```
nbuonin: Another suggestion: If `assets` is always at least an empty array, it might be nice to use `reduce` here. That way you can skip using `if`, and use 'All Class Members' object as an initial value.

media/js/src/AssetFilter.jsx
```diff
@@ -87,18 +149,25 @@ export default class AssetFilter extends React.Component {
                             </div>
                             <Select
                                 className="react-select"
-                                onChange={this.handleChange} options={[]} />
+                                isMulti
+                                onChange={this.handleChange}
```
nbuonin: I think you removed `handleChange`

media/js/src/AssetFilter.jsx
```diff
@@ -50,13 +113,14 @@ export default class AssetFilter extends React.Component {
                             <input
                                 type="text" name="title"
                                 className="form-control"
-                                onChange={this.props.handleTitleFilterChange}
-                                defaultValue={this.state.titleFilter}
                                 placeholder="Title of items and selections"
+                                onChange={this.props.handleTitleFilterChange}
+                                onKeyDown={this.handleTitleSearch.bind(this)}
```
nbuonin: Maybe bind in the constructor for consistency

media/js/src/AssetFilter.jsx
nikolas: This does need to be updated whenever the props are updated, though.. moving this to the constructor causes the options to not populate, since the assets are loaded over XHR - they start out as null. It does seem like a lot of work to do on each render - i will work on cleaning this up.

media/js/src/AssetFilter.jsx
nikolas: Yeah... i made this in the most basic way possible, just using for loops. I'll use `reduce` instead.

PR #2304, opened 2020-01-15
Refactor owner options population to use reduce()

nikolas:
And some other minor fixes.

### DIFF
--- a/media/js/src/AssetFilter.jsx
+++ b/media/js/src/AssetFilter.jsx
@@ -5,12 +5,14 @@ import Select from 'react-select';
 export default class AssetFilter extends React.Component {
     constructor(props) {
         super(props);
-        this.state = {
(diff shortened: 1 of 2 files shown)


media/js/src/AssetFilter.jsx
```diff
@@ -5,54 +5,64 @@ import Select from 'react-select';
...
+            label: 'All Class Members'
+        };
     }
     handleTitleSearch(e) {
         if (e.key === 'Enter') {
             this.props.handleTitleFilterSearch();
         }
     }
     handleOwnerChange(e) {
-        this.setState({selectedOwner: e.value});
```
nbuonin: Sorry to be a PITA, but did you mean to pull this?

nikolas: yes - i'm no longer using this state value.

PR #2342, opened 2020-02-13
Implement pagination in react collection view

nikolas:
Still TODO: the API needs to return some metadata about the total number
of items if we want to set up the page numbers properly.


sdreher:
re: total number of items, [at this point in the AssetCollectionView](https://github.com/ccnmtl/mediathread/blob/master/mediathread/assetmgr/views.py#L1120) the entire result set is available. Stuffing `assets.count()` into the `ctx` should get you what you need.

PR #2378, opened 2020-03-10
Add bootstrap alert on selection creation

nikolas:


media/js/src/AssetDetail.jsx
```diff
@@ -67,7 +73,10 @@ export default class AssetDetail extends React.Component {
                 end: 99
             }
         }).then(function() {
-            console.log('selection created');
+            me.setState({
+                createdSelectionTitle: selectionTitle,
+                showCreatedDialog: true
```
sdreher: Noting that the Create Dialog is not closing when I click "Save"...not sure if that's completely implemented yet?

nikolas: good point, i need to implement that.

sdreher: Great, I'll merge this PR then.

PR #2385, opened 2020-03-17
Add video selection playback - MEDT-3501

nikolas:
There's still an issue where the video doesn't work until you
actually hit "play" on it. I'm having trouble triggering the video
play with the react-player api. But once it's been played once, the
selections can be played.

### DIFF
--- a/media/js/src/AssetDetail.jsx
+++ b/media/js/src/AssetDetail.jsx
@@ -29,7 +29,13 @@ export default class AssetDetail extends React.Component {
             selectionLayer: new VectorLayer({
                 source: new VectorSource()
             }),
-            selectedSelection: null,
+
+            // The player seems to work better when it's loaded initially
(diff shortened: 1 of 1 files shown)


media/js/src/AssetDetail.jsx
```diff
@@ -29,7 +29,12 @@ export default class AssetDetail extends React.Component {
             selectionLayer: new VectorLayer({
                 source: new VectorSource()
             }),
-            selectedSelection: null,
+
+            // The player seems to work better when it's loaded initially
+            // as 'playing', and then paused immediately onReady.
+            // https://github.com/CookPete/react-player/issues/536#issuecomment-453869837
+            playing: true,
```
sdreher: In my dev env, when I click the video link from the collection, the video plays and does not pause.

nikolas: This has been addressed.

PR #2395, opened 2020-03-31
Add SherdNote api with validation - MEDT-3500

nikolas:
Using django-rest-framework.

### DIFF
--- a/mediathread/djangosherd/apiviews.py
+++ b/mediathread/djangosherd/apiviews.py
@@ -1,0 +1,38 @@
+from django.shortcuts import get_object_or_404
+from rest_framework import status
+from rest_framework.generics import CreateAPIView
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from mediathread.assetmgr.models import Asset
+from mediathread.djangosherd.serializers import SherdNoteSerializer
+from mediathread.djangosherd.permissions import AssetIsVisible, AssetInMyCourse
+from mediathread.sequence.apiviews import CsrfExemptSessionAuthentication
+
+
+class SherdNoteCreate(CreateAPIView):
+    authentication_classes = [CsrfExemptSessionAuthentication]
+    permission_classes = [IsAuthenticated, AssetIsVisible, AssetInMyCourse]
+    serializer_class = SherdNoteSerializer
+
+    def post(self, request, *args, **kwargs):
+        """
+        Create a new annotation.
+        """
+        asset_id = kwargs.get('asset_id')
+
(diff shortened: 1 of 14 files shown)


mediathread/djangosherd/apiviews.py
```diff
@@ -0,0 +1,26 @@
...
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from mediathread.djangosherd.serializers import SherdNoteSerializer
+from mediathread.sequence.apiviews import CsrfExemptSessionAuthentication
+
+
+class SherdNoteCreate(CreateAPIView):
+    authentication_classes = [CsrfExemptSessionAuthentication]
+    permission_classes = [IsAuthenticated]
```
sdreher: Additional permissions are needed to ensure the user can only create an annotation on an asset that is visible to them. We need a check that the asset is within a course where the user is a member. And, that the asset can be edited based on the course permissions.

nikolas: Thanks for listing these out! I will add these tomorrow.

mediathread/djangosherd/permissions.py
```diff
@@ -0,0 +1,38 @@
...
+            raise PermissionDenied
+
+        return True
+
+
+class AssetInMyCourse(BasePermission):
+    message = 'You can only update assets in your courses.'
+
+    def has_object_permission(self, request, view, obj):
+        # Am I the owner of this asset?
```
sdreher, line 31: This first check duplicates `AssetIsVisible`.  I don't think it is necessary here.

mediathread/djangosherd/permissions.py
```diff
@@ -0,0 +1,38 @@
...
+        if obj.author == request.user:
+            return True
+
+        if cached_course_is_faculty(obj.course, request.user):
+            return True
+
+        if not all_items_are_visible(request.course):
```
sdreher, line 21: ...should this check use `obj.course` for consistency?

mediathread/djangosherd/serializers.py
```diff
@@ -15,11 +15,73 @@ class Meta:
...
+
+        if not range1:
+            raise serializers.ValidationError({
+                'range1': 'This field is required.'
+            })
+
+        if not range2:
+            raise serializers.ValidationError({
+                'range2': 'This field is required.'
+            })
```
sdreher, line 74: Is a check needed here to ensure range1 < range2?

nikolas: Yes, I just added that.

PR #2399, opened 2020-04-01
Add SameSite=None flag for session cookies

nikolas:
Chrome 80 needs this flag set on our session cookie, in order to use it
for media collection.

* https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-SESSION_COOKIE_SAMESITE
* https://www.chromestatus.com/feature/5088147346030592
* https://www.chromestatus.com/feature/5633521622188032

### DIFF
--- a/mediathread/settings_shared.py
+++ b/mediathread/settings_shared.py
@@ -40,6 +40,9 @@ TEMPLATES[0]['OPTIONS']['context_processors'].append(  # noqa
     'mediathread.main.views.django_settings'
 )
 
+MIDDLEWARE = ['django_cookies_samesite.middleware.CookiesSameSite'] + \
+    MIDDLEWARE  # noqa
+
 MIDDLEWARE += [  # noqa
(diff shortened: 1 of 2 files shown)


mediathread/settings_shared.py
```diff
@@ -40,6 +40,9 @@
     'mediathread.main.views.django_settings'
 )
 
+MIDDLEWARE = ['django_cookies_samesite.middleware.CookiesSameSite'] + \
```
sdreher, line 43: Just a quick question - does this middleware need to be first in the array? Could it be added to the array at Ln 46?

nikolas: Yes, it does need to be first (https://github.com/jotes/django-cookies-samesite#quickstart) - I first tested this by putting it in the array below, but then realized that's just appending to what's already in ccnmtlsettings, and that's why it wasn't working!

If you have an idea of a nicer way to prepend this to the array, that could work! I'll also add a comment here that this needs to be the first middleware.

PR #2408, opened 2020-04-07
Clean up addressable courses

nikolas:
This pull request adds the course ID to many ajax requests in both the new and legacy JS code. The course session cookie is still set - this will still be used by the Chrome extension. I've added a note for that.

I've also removed the addressable_courses feature flag in this PR. We don't need a feature flag for this anymore - we can assume it's always true.


mediathread/urls.py
```diff
@@ -172,8 +172,8 @@
          name='affil_activate'),
 
     # New course-aware routes
-    path('course/<int:course_pk>/react/asset/<int:pk>/',
-         ReactAssetDetailView.as_view(),
+    path('course/<int:course_pk>/asset/<int:pk>/',
```
sdreher, line 175: Looking through this a bit...is there an *easy* way to preserve old functionality, e.g. like the "oldhome" setup. Largely as reference for QA efforts.

nikolas:
@sdreher this has been updated and cleaned up. It's ready for re-review.

Unit tests are passing, and there are just 2 cypress failures that may be related to what you're working on:

```
  1) Publish To World Public Composition
       enables publish to world in the course & creates project:
     AssertionError: Timed out retrying: Expected to find element: `.participants_chosen`, but never found it.

  1) Instructor Edits the Item Metadata
       should create a composition as an Instructor:
     AssertionError: Timed out retrying: expected '[ <li.annotation-global-body.meta>, 2 more... ]' to contain 'Here are my notes'
```

sdreher:
I'll review & take a look at the Cypress failures.

nikolas:
It looks like now only the publishtoworld.composition test is failing!

PR #2445, opened 2020-04-20
Handle html entities in asset titles

nikolas:
Unfortunately we need to use this trick to render HTML entities without
escaping them:
https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

Screen Shot 2020-04-20 at 3 58 47 PM

### DIFF
--- a/media/js/src/GridAsset.jsx
+++ b/media/js/src/GridAsset.jsx
@@ -219,8 +219,10 @@ export default class GridAsset extends React.Component {
                                 this.props.toggleAssetView.bind(
                                     this, this.props.asset)}
                             href={assetLink}
-                            title={this.props.asset.title}>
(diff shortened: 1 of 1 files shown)


media/js/src/GridAsset.jsx
```diff
@@ -219,8 +219,10 @@ export default class GridAsset extends React.Component {
                                 this.props.toggleAssetView.bind(
                                     this, this.props.asset)}
                             href={assetLink}
-                            title={this.props.asset.title}>
-                            {this.props.asset.title}
+                            title={this.props.asset.title}
+                            dangerouslySetInnerHTML={{
```
sdreher, line 223: ☠️☠️ ☠️ ☠️ ☠️ ☠️ ☠️

PR #2463, opened 2020-04-28
List View Improvements: add selection count column, fix sorting

nikolas:
Use database sort to sort assets. Use title for default asset sort. The
"assets" variable was getting passed around in assetmgr/views.py, but
ultimately ignored by the AssetResource.render_list() function. This
function can actually use the list of assets and preserve ordering.

This fixes the bugs we were seeing with asset ordering. And all the new
sorting code I added turned out to be unnecessary!

I realized that unit tests were only failing on python 2.7, because Python dictionary ordering behavior wasn't always defined clearly. I've disable these tests because they're not useful anymore: we're only running on python 3 on production. The lettuce tests on python 2 are still being run.

### DIFF
--- a/mediathread/assetmgr/api.py
+++ b/mediathread/assetmgr/api.py
@@ -1,13 +1,86 @@
 # pylint: disable-msg=R0904
-from django.urls import reverse
 import waffle
+import json
+import time
+from django.urls import reverse
 from mediathread.api import ClassLevelAuthentication, UserResource
 from mediathread.assetmgr.models import Asset, Source
 from mediathread.djangosherd.api import SherdNoteResource
 from tastypie import fields
 from tastypie.resources import ModelResource
-import json
-import time
+
+
+def sort_by_title(x):
+    return x.get('title', '').strip().lower()
+
+
+def sort_by_selections(x):
+    count = x.get('annotation_count', 0)
+    try:
+        return int(count)
+    except ValueError:
+        return 0
+
+
+def sort_by_author(x):
+    return x.get('author').get('public_name')
+
+
+def sort_by_default(order_by, x):
+    return x.get(order_by, None)
+
+
+def get_sort_function(order_by):
+    if order_by == 'title':
+        return sort_by_title
+    elif order_by == 'selections':
+        return sort_by_selections
+    elif order_by == 'author':
+        return sort_by_author
+
+    return sort_by_default(order_by)
+
+
+def sort_assets(assets, ordering):
+    order_by = ordering.get('order_by')
+    sort_func = get_sort_function(order_by)
+
+    return sorted(
+        assets,
+        key=sort_func,
+        reverse=ordering.get('reverse'))
+
+
+def get_ordering(request):
+    """
+    Get the ordering from a request object
+
+    Returns a dict with order_by (field name) and reverse (asc or
+    desc) values.
+    """
+    order_by = request.GET.get('order_by', 'title')
+    order_reverse = False
+    if order_by.startswith('-'):
+        order_by = order_by[1:]
+        order_reverse = True
+
+    return {
+        'order_by': order_by,
+        'reverse': order_reverse,
+    }
+
+
+def add_note_ctx_to_json(note_ctx, the_json):
+    if note_ctx['is_global_annotation']:
+        the_json['global_annotation'] = note_ctx
+
+        the_json['global_annotation_analysis'] = (
+            len(note_ctx['vocabulary']) > 0 or
+            len(note_ctx['metadata']['body']) > 0 or
+            len(note_ctx['metadata']['tags']) > 0)
+    else:
+        the_json['annotations'].append(note_ctx)
+    return the_json
 
 
 class AssetResource(ModelResource):
@@ -146,9 +219,6 @@ class AssetResource(ModelResource):
         note_resource = SherdNoteResource()
         ctx = {}
 
-        # Sort notes with tastypie's built-in ordering
-        notes = self.apply_sorting(notes, request.GET.dict())
-
         for note in notes.all():
             try:
                 note.asset.primary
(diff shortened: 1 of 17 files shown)


mediathread/assetmgr/api.py
```diff
@@ -147,7 +221,13 @@ def render_list(self, request, record_owner, record_viewer, assets, notes):
...
 
         # Sort notes with tastypie's built-in ordering
-        notes = self.apply_sorting(notes, request.GET.dict())
+        note_sort = request.GET.dict()
+
+        # SherdNote doesn't have selections to sort on
+        if re.match(r'^-?selections$', note_sort.get('order_by', '')):
+            note_sort['order_by'] = 'title'
+
+        notes = self.apply_sorting(notes, note_sort)
```
sdreher, line 230: The [AssetCollectionView](https://github.com/ccnmtl/mediathread/blob/master/mediathread/assetmgr/views.py#L1154) optimizes the render by paginating (slicing) the queryset before [it renders it out](https://github.com/ccnmtl/mediathread/blob/master/mediathread/assetmgr/views.py#L1159). (Rendering being expensive). Placing this sorting logic results in sorted pages, but not a full queryset sort. This logic needs to be moved up a level into the view.

mediathread/assetmgr/api.py
```diff
@@ -1,13 +1,87 @@
...
+        order_reverse = True
+
+    return {
+        'order_by': order_by,
+        'reverse': order_reverse,
+    }
+
+
+def add_note_ctx_to_json(note_ctx, the_json):
+    if note_ctx['is_global_annotation']:
```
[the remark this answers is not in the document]
nikolas: Ah okay, cool i'll remember that. This function isn't actually touched by this PR, I was just moving all the functions in this file above the API view class.

mediathread/assetmgr/api.py
```diff
@@ -1,13 +1,87 @@
...
+    order_by = ordering.get('order_by')
+    sort_func = get_sort_function(order_by)
+
+    return sorted(
+        assets,
+        key=sort_func,
+        reverse=ordering.get('reverse'))
+
+
+def get_ordering(request):
```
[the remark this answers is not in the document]
nikolas: No, I'll add them to this PR.

mediathread/assetmgr/api.py
nikolas: This part does seem to be working, sorting the annotations tied to each asset returned. I don't think it's necessary to sort the annotations here, so I'll just remove this part, for clarity.

nikolas: I see what you're talking about though - my asset sort is not working on the entire set of assets still. I'll fix that up with your tips.

mediathread/assetmgr/views.py
```diff
@@ -1118,25 +1117,31 @@ def add_metadata(self, request, assets):
...
+                request.course,
+                self.record_owner)
         else:
             assets = Asset.objects.by_course(request.course)
 
         (assets, notes) = self.visible_assets_and_notes(request, assets)
 
+        # Order by title, by default.
+        ordering = request.GET.get('order_by', 'title')
+        if ordering == 'title' or ordering == 'author':
```
sdreher, line 1140: I'm seeing sort inconsistencies when the sort order is `-author` or `-title`. Likely lowercase needs to be handled for those cases as well.

nikolas: oh yeah, of course

nikolas:
I've addressed the default sort issue by adding the defaultSortField option to the react-data-table. So, it's now clear that it's sorting by title, ascending by default. And clicking that once will reverse that.

mediathread/main/tests/test_views.py
```diff
@@ -454,7 +454,7 @@ def test_migrate_materials_sample_course(self):
 
         the_json = json.loads(response.content)
         self.assertEquals(the_json['course']['title'], 'Sample Course')
-        self.assertEquals(len(the_json['assets']), 1)
+        self.assertEquals(len(the_json['assets']), 2)
```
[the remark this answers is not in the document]
nikolas: This test changed when I made the change from populating the asset context from `notes` to `assets`. I'm looking further into this to find out exactly why we're now getting two assets here instead of one..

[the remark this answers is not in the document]
nikolas: Okay, see commit 89f1d7794266be8ff1c31701a56f7d9188fd7b56. I've changed the test here back to its original state.

In the MigrateMaterialsView, notes was excluding the flv_pseudo type, but not assets, which is now used as well. So, I've added an exclude case here. It seems right to me, but I'm not really familiar with all these types, why flv_pseudo is getting excluded here in the first place.

nikolas:
User name display is fixed with the latest commit here.

For the modified date sorting problems... yeah, looks like it's due to this code path that updates the Asset's modified time if it has a note with a more recent modified time. My new sort backend doesn't know about this new value: https://github.com/ccnmtl/mediathread/blob/master/mediathread/assetmgr/api.py#L132

I'm going to have to think this over some more. So, I'll address that either before or after this pull request is merged.

sdreher:
In response to the modified date issue - as you've pointed out - adding a sort for the asset's note_set is tricky due to visible selections. Instead of `modified`, maybe the asset's `added` date would be more useful, and definitely more sortable.

mediathread/assetmgr/features/collection.feature
```diff
@@ -35,8 +35,8 @@ Feature: Collection View
         And the Collection panel has a "Mediathread: Introduction" item
 
         When I clear all tags
-        When I select "Student One" as the owner
-        Then the owner is "Student One" in the Collection column
+        When I select "One, Student" as the owner
```
sdreher, line 38: Oi! I didn't think about these tests when we talked about changing the display.

nikolas:
Alright thanks for the suggestion. I added a commit here to sort on Asset.added instead of Asset.modified.

mediathread/assetmgr/features/collection.feature
nikolas: Yeah, changing to last name, first name for name display will make some things feel less natural. This test stuff doesn't matter too much though.. "Student One", "Instructor One", etc, are only used in testing.

PR #2545, opened 2020-05-27
Add new asset detail tab layout - MEDT-3508

nikolas:


nikolas:
there's a problem with loading the Video asset page here that I need to fix.

nikolas:
I downgraded `react-player`, and made some updates to the create selection component, which fixes that problem.

PR #2570, opened 2020-06-01
Improve items tabs

nikolas:
Implement more of Marc's work here.


media/js/src/assetDetail/AssetDetail.jsx
```diff
@@ -288,6 +288,39 @@ export default class AssetDetail extends React.Component {
 
                     <div className="col-sm-6">
                         {media}
+                        {this.state.tab === 'viewItem' && (
```
sdreher, line 291: I'd like to strongly lobby for this section to be placed under the "View the Original Item" tab. It feels unmoored below the video. I'll merge this, and we can discuss.

PR #2594, opened 2020-06-08
Composition tests

ndittren:


cypress/integration/composition/composition.creation_Instructor.js
```diff
@@ -0,0 +1,69 @@
...
+
+    beforeEach(() => {
+      cy.login('instructor_one', 'test');
+      cy.visit('/course/1/', { timeout: 80000 });
+    });
+
+    it('should check composition panel features', () => {
+      cy.get('#homepage-create-menu').should('exist').click();
+      cy.get('#create-project-menu input[type="submit"]')
+        .eq(1).contains('Create Composition').click();
```
nbuonin, line 19: You may want to think about refactoring this a bit. `.contains` will match the first text it finds, and in this case it matches "Create Composition Assignment" and "Create Composition".
Two you can do:
* Where possible, add `data-cy` attributes to the template rather than relying on css selectors. That way the markup can be refactored without having to update the tests. It will also make it easier to zero in on elements for assertions. See this: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements
* Break up this particular line into two, the first one for the assertion, the second for the click event

ndittren: Will do! I was trying to avoid changing the template as much as possible but this one is a little tricky to add which is why I added the `.eq(1)` But maybe I can use `to.eq` directly on the string? I'll definitely break it up!

nikolas:
you probably want to rebase this branch on the latest `master` - there was a cypress update I just merged here: https://github.com/ccnmtl/mediathread/pull/2600

PR #2598, opened 2020-06-08
Hide pagination on asset detail page

nikolas:


media/js/src/CollectionTab.jsx
```diff
@@ -114,6 +114,7 @@ export default class CollectionTab extends React.Component {
                 <AssetFilter
                     assets={this.props.assets}
                     assetCount={this.props.assetCount}
+                    hidePagination={!!this.state.selectedAsset}
```
sdreher, line 117: I'm not familiar with this syntax...double negation?

nikolas: Yeah, this turns a nullable object into a boolean. If there is a selectedAsset, this will be `true`, otherwise it will be `false`.

PR #2601, opened 2020-06-09
Use new course view by default

nikolas:
Remove the /react/ path.

This includes a bunch of other test-related fixes, too.


sdreher:
I'm concerned about commenting out these tests. Could we mock the project creation flow in Cypress? I know we've taken this path for programmatic login.

nikolas:
Yeah, I think we might be able to mock the POST request in cypress.. I'll update this branch. https://docs.cypress.io/api/commands/request.html#Syntax

nikolas:
This branch is now passing, though the cypress tests seem not quite stable (using some half-second waits). We'll improve these as we go.

cypress/integration/2.composition/composition.creation_instructor.js
```diff
@@ -7,19 +7,24 @@ Cypress.on('uncaught:exception', (err, runnable) => {
...
-            .eq(1).click();
+
+        cy.visit('/course/1/project/create/', {
+            method: 'POST',
+            body: {
+                project_type: 'composition'
+            }
+        });
+
+        cy.wait(500);
```
sdreher, line 26: Does Cypress have a `wait until` pattern?

cypress/integration/2.composition/composition.creation_instructor.js
```diff
@@ -7,19 +7,24 @@ Cypress.on('uncaught:exception', (err, runnable) => {
...
 
     it('should create a composition as an Instructor', () => {
         cy.log('should check composition panel features');
-        cy.get('#homepage-create-menu').should('exist').click();
-        cy.get('#create-project-menu input[type="submit"]')
-            .eq(1).contains('Create Composition')
-        cy.get('#create-project-menu input[type="submit"]')
-            .eq(1).click();
+
+        cy.visit('/course/1/project/create/', {
```
sdreher, line 19: Nice quick fix for the composition create problem.

mediathread/settings_shared.py
```diff
@@ -219,8 +219,10 @@ def default_url_processor(url, label=None, request=None):
 if 'test' in sys.argv or \
    'jenkins' in sys.argv or \
    'integrationserver' in sys.argv:
-    DEBUG = False
+    DEBUG = True
```
sdreher, line 222: Why switch this to `True`?

mediathread/settings_test.py
```diff
@@ -42,6 +42,8 @@
     'django_statsd.middleware.GraphiteMiddleware')
 MIDDLEWARE.remove(
     'impersonate.middleware.ImpersonateMiddleware')
+MIDDLEWARE.remove(
```
sdreher, line 45: Is the debug toolbar needed here?

mediathread/settings_shared.py
nikolas: So when I get 500 errors in cypress, the error is displayed fully.

mediathread/settings_test.py
nikolas: No - these settings are used for lettuce tests. I was just trying to simplify things.

cypress/integration/2.composition/composition.creation_instructor.js
nikolas: Yeah - you can use `wait` to wait for specific XHR requests to come through.... that's just, more work to set up that doesn't necessarily need to be done in this pull request: https://docs.cypress.io/api/commands/wait.html#Alias

PR #2625, opened 2020-06-17
Quick edit tests

ndittren:
*Cypress tests for Quick Edit feature
*Small tweaks to Composition test

### DIFF
--- a/cypress/integration/4.quickedit/create.selection.js
+++ b/cypress/integration/4.quickedit/create.selection.js
@@ -9,8 +9,9 @@ Cypress.on('uncaught:exception', (err, runnable) => {
 describe('Instructor creates a selection', () => {
     before(() => {
         cy.login('instructor_one', 'test');
-        cy.visit('/course/1/');
-        cy.wait();
+        cy.visit('/course/1/').should((win) => {
+            expect(win.MediaThread).to.exist;
+        });
     });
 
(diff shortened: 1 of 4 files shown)


cypress/integration/4.quickedit/create.selection.js
```diff
@@ -0,0 +1,62 @@
...
+    // returning false here prevents Cypress from
+    // failing the test
+    return false;
+});
+
+describe('Instructor creates a selection', () => {
+    before(() => {
+        cy.login('instructor_one', 'test');
+        cy.visit('/course/1/');
+        cy.wait();
```
nbuonin, line 13: One thing you could try here is chaining `should` to `visit`, like:
```
cy.visit('/course/1/').should(() => { expect(...)});
```
The `expect` syntax comes from Chai, but Cypress has some examples here: https://docs.cypress.io/guides/references/assertions.html#Chai
In any case, the thing that helped me understand this better was that if some action had side effects on the page, like an AJAX request, then use this chained syntax.

nbuonin: Here's another example from the docs: https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Asserting-in-English
```
cy.request('/users/1').its('body').should('deep.eq', { name: 'Jane' })
```

PR #2646, opened 2020-06-23
:arrow_up: Django 2.2

nikolas:
* Update a few package dependencies that were held back by django 1.11.
* Remove virtualenv / update django.mk


mediathread/djangosherd/models.py
```diff
@@ -327,8 +327,11 @@ class SherdNote(Annotation):
     objects = SherdNoteManager()
 
     title = models.CharField(blank=True, max_length=1024, null=True)
-    asset = models.ForeignKey(Asset, related_name="sherdnote_set")
-    author = models.ForeignKey(User, null=True, blank=True)
+    asset = models.ForeignKey(
+        Asset, related_name="sherdnote_set",
+        on_delete=models.CASCADE)
+    author = models.ForeignKey(
```
sdreher, line 333: Wondering about PROTECT here? If the author is deleted, I think the annotations should cascade.

mediathread/djangosherd/models.py
```diff
@@ -415,14 +418,16 @@ class DiscussionIndex(models.Model):
...
-    asset = models.ForeignKey(Asset, null=True,
-                              related_name="discussion_references")
+    asset = models.ForeignKey(
+        Asset, null=True,
+        related_name="discussion_references",
+        on_delete=models.CASCADE)
 
     # just for use-case #3
-    comment = models.ForeignKey(Comment, null=True)
+    comment = models.ForeignKey(Comment, null=True, on_delete=models.PROTECT)
```
sdreher, line 430: Also wondering about PROTECT here...If a comment is deleted, it makes sense to me that the DiscussionIndex would also be deleted.

mediathread/main/tests/test_views.py
```diff
@@ -1133,8 +1134,10 @@ def test_email_invite_invalid_email(self):
             in response.cookies['messages'].value)
 
     def test_accept_invite_invalid_uuid(self):
-        # no uuid
-        url = reverse('course-invite-accept', kwargs={'uidb64': None})
+        # random new uuid
+        url = reverse(
```
sdreher, line 1138: Why change the test here from None to random? Are the kwargs required in Django 2.2?

mediathread/settings_shared.py
```diff
@@ -184,7 +184,7 @@ def no_reject(request, reason):
 CORS_ALLOW_METHODS = ('GET',)
 CORS_ALLOW_CREDENTIALS = True
 
-SESSION_COOKIE_SAMESITE = 'None'
+DCS_SESSION_COOKIE_SAMESITE = 'None'
```
sdreher, line 187: What is `DCS_SESSION_COOKIE_SAMESITE`?

nikolas: DCS stands for django-cookies-samesite. This is the module we're using to add SAMESITE headers to the session cookie. Django introduced support for this in 2.0, but it's not complete. We can use django's functionality once we're on 3.1, which includes this fix: https://github.com/django/django/commit/b33bfc383935cd26e19a2cf71d066ac6edd1425f

For now, we'll need to continue to use the django-cookies-samesite package, which requires this setting name change to not conflict with Django's setting. Described here: https://pypi.org/project/django-cookies-samesite/

mediathread/djangosherd/models.py
nikolas: Yeah, I agree. I'll update that.

mediathread/discussions/views.py
```diff
@@ -129,20 +132,23 @@ def discussion_create(request):
 
 class DiscussionDeleteView(LoggedInFacultyMixin, View):
     def post(self, request, discussion_id):
-        root_comment = get_object_or_404(ThreadedComment, pk=discussion_id)
+        discussion = get_object_or_404(DiscussionIndex, pk=discussion_id)
```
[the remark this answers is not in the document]
nikolas: I changed this because I thought it didn't make sense how we were looking up a ThreadedComment by its discussion ID. The tests were failing here, and this resolved it. But, I this will also be resolved by changing the comment PROTECT on_delete attribute to CASCADE, like you mentioned above. So, I'll revert this to how it was originally working.

mediathread/djangosherd/models.py
nikolas: Yeah - this has been updated.

mediathread/discussions/views.py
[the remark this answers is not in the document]
sdreher: DiscussionIndex is this other "thing" that was setup to support the course feed...which we don't really use anymore. I haven't looked at it in years, but it was basically constructed as a denormalized data view. It's unrelated to ThreadedComments. The `discussion_id` here references the parent-level ThreadedComment.

mediathread/settings_shared.py
sdreher: Great. Thanks for the explanation.

mediathread/main/tests/test_views.py
nikolas: Basically - yeah. Django 2.2 causes the request with `'uidb64': None` to fail with an exception. We could catch that exception if we want to test the `/accounts/invite/accept/None/` route... but that seemed sort of contrived, and what real situation would actually cause that request to happen?

So, instead I'm just testing with a random, invalid UUID here. I suppose we could also test that just getting the URL with a random non-uuid string fails as gracefully as possible, too.

PR #2668, opened 2020-06-26
Asset detail dom cleanup

nikolas:


media/js/src/assetDetail/ViewSelections.jsx
```diff
@@ -69,10 +69,6 @@ export default class ViewSelections extends React.Component {
 
         return (
             <React.Fragment>
-                <h2>
-                    Selections from <a href="#">{this.props.asset.title}</a>
```
sdreher, line 73: This should keep `Selections` and remove the rest.

media/js/src/assetDetail/CreateSelection.jsx
```diff
@@ -11,8 +11,6 @@ export default class CreateSelection extends React.Component {
     render() {
         return (
             <React.Fragment>
-                <h3>Additional Selection Information</h3>
```
sdreher, line 14: We wanted to keep this header and remove the secondary one. "Add Selection Details"

PR #2677, opened 2020-06-27
Change react-player width/height to 100%

nikolas:
width: auto seems to be hiding the video for me.. for some reason I
didn't notice this before.


sdreher:
Nice. This works for me.

PR #2757, opened 2020-07-13
Assignments List: Add a logical sort order for student view

sdreher:
* the student's response submitted date asc, with nulls first (indicating no response)
* the timedelta between today and the due_date desc with nulls last

To accomplish this, I took advantage of Django's Subquery filter and advanced annotate features. This chokes sqlite3 unfortunately, so I'm also switching the tests to run on postgres. @nikolas - I'd like to verify you're okay with this switch.


nikolas:
This looks great, and yeah - if we can get the tests running on postgres instead of sqlite, that would be great! I'm all for that change.. whatever works. It's just a matter of making sure that works on travis and jenkins.

PR #2779, opened 2020-07-17
add to collection button

ndittren:
* added a button that leads to the upload/import page
* needs new bundle.js

### DIFF
--- a/media/js/src/CollectionTab.jsx
+++ b/media/js/src/CollectionTab.jsx
@@ -140,11 +140,10 @@ export default class CollectionTab extends React.Component {
 
                     <h1 className="page-title">Collection</h1>
 
-                    <div className="btn btn-outline-secondary" aria-label="Add to Collection">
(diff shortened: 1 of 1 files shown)


media/js/src/CollectionTab.jsx
```diff
@@ -135,8 +135,16 @@ export default class CollectionTab extends React.Component {
 
         return (
             <div role="tabpanel">
-                <h1 className="page-title">Collection</h1>
 
+                <div className="d-flex justify-content-between align-items-center flex-wrap">
+
+                    <h1 className="page-title">Collection</h1>
+
+                    <div className="btn btn-outline-secondary" aria-label="Add to Collection">
```
sdreher, line 143: The `className="btn btn-outline-secondary" aria-label="Add to Collection"` should go on the anchor tag below.

ndittren: Ah yes!

PR #2786, opened 2020-07-20
Composition assignment wizard

sdreher:


media/css/mediathread_new.css
```diff
@@ -52,6 +52,10 @@ h1.page-title {
     padding: .25em 0em .25em 0em;
 }
 
+.hidden {
```
nikolas, line 55: There is also the `.d-none` class in Bootstrap 4: https://getbootstrap.com/docs/4.5/utilities/display/#hiding-elements

sdreher: Good note. Yes - I had changed it to `d-none`, then had to change it back as "hidden" is still used in the Selection Assignment wizard. I have a note to migrate.

PR #2792, opened 2020-07-21
Add Delete selection button, in Edit form

nikolas:
Most of this functionality was already implemented, this just hooks it
up to the Edit form.

### DIFF
--- a/media/js/src/forms/EditSelectionForm.jsx
+++ b/media/js/src/forms/EditSelectionForm.jsx
@@ -155,13 +155,22 @@ export default class EditSelectionForm extends React.Component {
                 </Form.Group>
 
                 <Button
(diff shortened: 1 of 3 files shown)


media/js/src/forms/EditSelectionForm.jsx
```diff
@@ -162,6 +162,14 @@ export default class EditSelectionForm extends React.Component {
                 <Button type="submit" size="sm">
                     Save
                 </Button>
+                <div className="float-right">
```
sdreher, line 165: Can the Delete button be left, with Cancel, Save on the right? I think the natural instinct is to click the right-hand button. Jira ticket has details + image.

nikolas:
I've updated the "Selection Deleted" alert with those changes.

PR #2796, opened 2020-07-21
Drawing improvements - MEDT-3596

nikolas:
* Disable panning and zooming when drawing
* Fix a problem with onClearActiveSelection, where it wasn't clearing
  everything
* Enable Clear button
* Only allow one shape per selection
* Update onCreateSelection to only save one feature. This fixes a bug
  where selections created in the new tool weren't displaying correctly,
  because the coordinates were nested incorrectly.
* Removed focus style from Shape button. This fixes the issue where you
  have to click twice on the image map after clicking the Shape button.

### DIFF
--- a/media/js/src/assetDetail/AssetDetail.jsx
+++ b/media/js/src/assetDetail/AssetDetail.jsx
@@ -15,6 +15,8 @@ import VectorLayer from 'ol/layer/Vector';
 import VectorSource from 'ol/source/Vector';
 import Projection from 'ol/proj/Projection';
 import Static from 'ol/source/ImageStatic';
+import {defaults as defaultControls} from 'ol/control';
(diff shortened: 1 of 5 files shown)


media/js/src/assetDetail/AssetDetail.jsx
```diff
@@ -581,6 +588,20 @@ export default class AssetDetail extends React.Component {
                 this.startButtonRef.current.focus();
             }
         }
+
+        if (prevState.isDrawing !== this.state.isDrawing) {
```
sdreher, line 592: I'm seeing this work the first time through, but then reappearing in subsequent "draw" events. Let's talk through the desired flow here tomorrow morning.

sdreher:
This is looking very good. Good call on removing the orange styling, it just wasn't working well. :clap:

PR #2799, opened 2020-07-22
Card active state with drop shadow

sdreher:


media/css/mediathread_new.css
```diff
@@ -4351,6 +4351,25 @@ ul.section-tabs>li>a.active {
...
+}
+
+.course-collection .card {
+    margin-bottom: .25em;
+}
+
+.course-collection .card.active {
+    box-shadow: 0 .5rem 1rem rgba(0,0,0,.15)!important;
+    border-radius: .25rem!important;
+    /*margin-bottom: 3rem!important;*/
```
nikolas, line 4366: This CSS should be cleaned up a bit. `!important` should really never be necessary.. if the selector is specific enough it will get the right style. And, the commented line removed.

sdreher:
Thanks for taking care of the .css files @nikolas. Very appreciated.

PR #2841, opened 2020-07-27
Filter improvements

nikolas:
* Reset current page to 1 on filter change
* Default asset filter to "Me"
* Fix some pagination problems
* Use state variables for filter values


sdreher:
looks good!

PR #2902, opened 2020-08-05
Alert messages

ndittren:


mediathread/templates/dashboard/base_course_dashboard.html
```diff
@@ -2,6 +2,19 @@
 
 {% block title %}&mdash; Manage Course {% endblock %}
 
+{% block messages %}
+    {% if messages %}
+        {% for message in messages %}
+        <div class="alert alert-warning alert-dismissible fade show" role="alert">
```
sdreher, line 8: This looks good. The only change -- I'd say that we should use "alert-info" rather than warning here as the messages are mostly confirming an action.

PR #2907, opened 2020-08-05
Disable csrf on /save/ route, for media collection

nikolas:


mediathread/assetmgr/views.py
```diff
@@ -160,6 +161,7 @@ def most_recent(request):
     return HttpResponseRedirect('/asset/' + asset_id + '/')
 
 
+@method_decorator(csrf_exempt, name='dispatch')
```
sdreher, line 164: Why do we need this exempted? I thought the react code is using the authed_fetch?

nikolas: This view gets used from the browser extension, not from within Mediathread. So.. maybe it's possible to communicate the csrf code to the extension, and use it there, but that will require some changes in the extension.

sdreher: Ah! makes sense. Maybe add a comment in there so we don't forget this?

PR #2909, opened 2020-08-05
Help accordion

ndittren:
* Help section turned into accordion for Migration
* Help section turned into accordion for Taxonomy


sdreher:
These look great!

PR #2918, opened 2020-08-07
Remove unused apache files

nikolas:


sdreher:
Some of our partners use Apache to serve their instances of Mediathread. Likely we should leave them.

PR #2938, opened 2020-08-11
Fix cypress tests on master by reverting JS build

nikolas:
Okay, so I guess this is the danger of travis not making a new JS build when running the tests! That should be easy to fix by adding a `npm run build` command to `.travis.yml`.

Revert 3d8f8ae9654a0f1a825e83ca695205de0a4943df

And then, i'll have to figure out which JS change caused these failures...


sdreher:
Good plan for the `npm run build`. I don't think we need to revert. This should hopefully resolve the issue: https://github.com/ccnmtl/mediathread/pulls

PR #2942, opened 2020-08-11
Scroll to selection's accordion on save/create

nikolas:


nikolas:
This is fixed. I had forgot about a `scrollIntoView` param I made for `openSelectionAccordionItem()` that I just wasn't using.

PR #2944, opened 2020-08-11
Hide Collection/Add to collection button when needed - MEDT-3706

nikolas:


nikolas:
@sdreher that problem is fixed now.

PR #2947, opened 2020-08-11
Implement edit image annotation shape + video selection

nikolas:
This includes a big refactoring: implement SelectionForm, which the
creation from and edit form both inherit.

### DIFF
--- a/media/js/src/assetDetail/AssetDetail.jsx
+++ b/media/js/src/assetDetail/AssetDetail.jsx
@@ -89,7 +89,8 @@ export default class AssetDetail extends React.Component {
 
             tab: 'viewSelections',
 
-            isDrawing: false
+            isDrawing: false,
+            isEditing: null
         };
 
         this.draw = null;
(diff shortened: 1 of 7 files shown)


media/js/src/assetDetail/AssetDetail.jsx
```diff
@@ -525,7 +566,7 @@ export default class AssetDetail extends React.Component {
                                     <div className="col-md-9">
                                         <div className="input-group">
 
-                                            {this.state.tab === 'createSelection' && (
+                                            {(this.state.tab === 'createSelection' || this.state.isEditing) && (
```
sdreher, line 569: The selection start & selection end buttons & timecode spinners show correctly in edit mode. But, the textual range display is also showing up. Looking at line 585, maybe a `!isEditing` needs to be added?

media/js/src/forms/SelectionForm.jsx
```diff
@@ -0,0 +1,228 @@
...
+                    !hasSelection(this.props.selectionSource)
+            ) || (
+                this.props.type === 'video' &&
+                    !this.props.selectionStartTime &&
+                    !this.props.selectionEndTime
+            )
+        ) {
+            e.preventDefault();
+            e.stopPropagation();
+            this.props.onShowValidationError(noSelectionError);
```
sdreher, line 83: For video editing, the code is falling into this path for valid selection start & end ranges. I'm not quickly seeing why this is the case.

nikolas: This is resolved - the video selection props weren't getting passed in to this new form.

nikolas:
These problems are fixed now.

PR #2952, opened 2020-08-11
Improve grid view - MEDT-3705

nikolas:
Screenshot_2020-08-11_18-40-16

I'm putting the "(1 of 2)" text before the title, because the CSS ellipsis functionality isn't advanced enough to determine available space, and append a string after the ellipsis. If we put this info at the end, it won't be visible to the user, unless we put this info in a separate element and do some sort of flexbox solution, which is definitely possible. But this works for now:

Screenshot_2020-08-11_18-41-15

### DIFF
--- a/media/js/src/GridAsset.jsx
+++ b/media/js/src/GridAsset.jsx
@@ -17,7 +17,9 @@ import Asset from './Asset';
 import {
     capitalizeFirstLetter, handleBrokenImage, getAssetUrl
 } from './utils';
-import {objectProportioned, displaySelection} from './openlayersUtils';
+import {
+    objectProportioned, displaySelection, resetMap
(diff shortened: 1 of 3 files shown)


media/js/src/GridAsset.jsx
```diff
@@ -69,8 +82,8 @@ export default class GridAsset extends React.Component {
         const assetUrl = getAssetUrl(this.props.asset.id);
 
         return (
-            <div className="col-sm-4">
-                <h5 className="text-nowrap text-truncate">
+            <div className="col-sm-4 border-right border-white mb-2">
```
sdreher, line 85: This is looking better spacing-wise, but we can probably bump it up just a shade more. We can talk tomorrow.

PR #2953, opened 2020-08-12
Revert "Collection: open the "add" accordion when collection is empty"

sdreher:
Reverts ccnmtl/mediathread#2949

There's an ugly bug with this logic. I'm going to revert and we can talk through how to do this properly. Likely React should handle this.


nikolas:
Yes, i noticed that! Well, it was actually helpful: it made me notice some wording changes in the browser extension area that I've made pull requests for.

PR #2955, opened 2020-08-12
Update browser extension requirements

nikolas:
The extension actually works with any browser based on Chromium (I use
it in Brave):
https://en.wikipedia.org/wiki/Chromium_(web_browser)#Browsers_based_on_Chromium

So, I don't want users of other browsers to feel left out here. I think
"a browser in the Chrome family" is succint enough, and still makes it
obvious for less technical users that they just need to use Chrome.

Changing this line:

    You must be using Chrome to install this extension.

to:

    You must be using a browser in the Chrome family to install this
    extension.

### DIFF
--- a/mediathread/templates/main/collection_add.html
+++ b/mediathread/templates/main/collection_add.html
@@ -15,7 +15,12 @@
                        <div class="card-body">
                            <h5 class="card-title text-center">Import Media</h5>
                            <p class="card-text">
-                               Install Mediathread’s Google Chrome extension to import video, audio, and images into this course from various sites across the web. You must be using Chrome to install this extension.
+                               Install Mediathread’s Google Chrome
+                               extension to import video, audio, and
+                               images into this course from various
+                               sites across the web. You must be using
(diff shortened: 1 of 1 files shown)


mediathread/templates/main/collection_add.html
```diff
@@ -16,7 +16,12 @@ <h4 class="text-center pt-3">Add To Collection</h4>
                        <div class="card-body">
                            <h5 class="card-title text-center">Import Media</h5>
                            <p class="card-text">
-                               Install Mediathread’s Google Chrome extension to import video, audio, and images into this course from various sites across the web. You must be using Chrome to install this extension.
+                               Install Mediathread’s Google Chrome
+                               extension to import video, audio, and
+                               images into this course from various
+                               sites across the web. You must be using
+                               a browser in the Chrome family to
+                               install this extension.
```
sdreher, line 24: Why not provide the wikipedia link here? Likely folks would find that helpful.

nikolas: good idea

PR #2966, opened 2020-08-13
Sync collection's assets state when leaving asset detail - MEDT-3713

nikolas:


media/js/src/CollectionTab.jsx
```diff
@@ -71,18 +71,24 @@ export default class CollectionTab extends React.Component {
...
+        // Always reload the assets when leaving asset defail view.
+        // The user might have made changes to the selections, and
+        // this needs to be up to date.
+
+        // Set assets to null, to display "Loading assets".
+        this.props.onUpdateAssets(null);
+
+        // Reload the assets.
+        const me = this;
+        getAssets().then(function(d) {
```
sdreher, line 83: After creating a selection and navigating back to the full collection, the collection refreshes to show all items rather than those from "Me". I'm also wondering if a full refresh is needed, or just a refresh for the asset that was being changed?

nikolas: Hmmm, it's true. Reloading all the assets was the simplest way to do this, but now that I've thought it out, I think it is possible to just setState with a new `assets` array, and include the updated state of the asset. This would avoid an extra ajax call.

sdreher: That might be a good approach...getting the assets is still a bit slow. Based on your comments, Is this PR ready to be merged, or should I wait?

nikolas: I will make that change here in this PR - hold on.

PR #2971, opened 2020-08-14
Redirect from old collection to new collection

sdreher:
If the user is attempting to access the non-ajax collection, asset or annotation views, send them over to the new view.

This also refactors the "most recent asset" view...which I think was used with the bookmarklet to navigate to the collection and show the last thing collected. @nikolas - is the extension still using this?


nikolas:
I don't think the extension is still using the "most recent asset" view.

PR #3016, opened 2020-08-24
Re-organize Item Metadata view - MEDT-3631

nikolas:
Screen Shot 2020-08-24 at 1 08 03 PM

### DIFF
--- a/media/js/src/assetDetail/ViewItem.jsx
+++ b/media/js/src/assetDetail/ViewItem.jsx
@@ -3,9 +3,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import Alert from 'react-bootstrap/Alert';
+import find from 'lodash/find';
 
-import {getAssetReferences} from '../utils';
+import {getAssetReferences, getTags, getTerms} from '../utils';
 
 export default class ViewItem extends React.Component {
     constructor(props) {
@@ -22,8 +22,6 @@ export default class ViewItem extends React.Component {
         });
     }
     render() {
-        const references = [];
-
         let authorId = null;
         if (this.props.asset && this.props.asset.author) {
             authorId = this.props.asset.author.id;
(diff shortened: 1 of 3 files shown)


media/js/src/assetDetail/ViewItem.jsx
```diff
@@ -34,34 +34,85 @@ export default class ViewItem extends React.Component {
...
+                description = desc[0];
+            }
+        }
+
         return (
             <div className="tab-content" id="pills-tabContent">
-                <table className="table mt-1">
+                <h3>Tags</h3>
+                <p>
+                    {tags}
```
sdreher, line 84: I'm seeing this print out as `[object Object]`

media/js/src/assetDetail/ViewItem.jsx
```diff
@@ -34,34 +34,85 @@ export default class ViewItem extends React.Component {
...
+                        <a href={reference.url}>{reference.title}</a>
+                    </li>
                 );
             });
         }
 
+        let tags = '(None)';
+        if (
+            this.props.asset &&
+                this.props.asset.global_annotation &&
```
sdreher, line 51: I'm not sure we want to pull the tags & terms off the global_annotation, as that idea is being discontinued. Can we show all the tags & terms from the visible selections on an item?

nikolas: Ah okay - yeah I can definitely pull tags off the item's selections. I did realize global_annotation is getting deprecated, I just didn't see anywhere else tags were present on the item.

sdreher: Tags & terms aren't present on the item itself. The `global_annotation` is a weird "special" annotation that's created to signify the item is within the user's collection. Tags, terms and notes for the "whole item" were kept on the `global_annotation`. Since we've deprecated the idea of adding tags, terms and notes to the item, we can kill the `global_annotation`, and rewrite the collection api to show any items a user owns, e.g. uploaded or imported, and any items a user has annotated. Future work to do.

nikolas:
I removed the reference to global_annotation, and fixed tag/term display to display off the asset's annotations.

media/js/src/assetDetail/ViewItem.jsx
```diff
@@ -34,34 +34,78 @@ export default class ViewItem extends React.Component {
...
+                <h3>Course References</h3>
+                <Alert variant="warning" show={
+                    this.state.references && this.state.references.length === 0
+                }>
+                    There are no references in this course.
+                </Alert>
+
+                {this.state.references && this.state.references.length > 0 && (
+                    <>
+                        <h4>Assignments</h4>
```
sdreher, line 98: References could be discussions or projects as well. I think we can pull this title.

media/js/src/assetDetail/ViewItem.jsx
```diff
@@ -34,34 +34,75 @@ export default class ViewItem extends React.Component {
...
+
+                <h3>Course Vocabulary (Terms)</h3>
+                <p>
+                    {terms}
+                </p>
+
+                <hr />
+
+                <h3>Course References</h3>
+                <Alert variant="warning" show={
```
sdreher, line 90: Tags & terms display plain text "(None)" when nothing exists. Can we make this consistent and pull the alert style here?

media/js/src/assetDetail/ViewItem.jsx
```diff
@@ -34,34 +34,75 @@ export default class ViewItem extends React.Component {
...
             <div className="tab-content" id="pills-tabContent">
-                <table className="table mt-1">
+                <h3>Tags</h3>
+                <p>
+                    {tags}
+                </p>
+
+                <hr />
+
+                <h3>Course Vocabulary (Terms)</h3>
```
sdreher, line 82: We always call this "terms" in the interface. Until the team as a whole renames this feature, can the title just say "Terms" for consistency?

media/js/src/assetDetail/ViewItem.jsx
nikolas: I don't know, I think that alert is a bit much. Honestly I think we should just hide any headers that don't have any elements. But, here's what it would look like with the alerts:

Screen Shot 2020-08-25 at 1 25 00 PM

nikolas: Oh maybe I misunderstood what you're saying here - should I make them all `(None)`?

sdreher: I'm not a big fan of the alerts either. I think make them all "None", or "There are no tags", "There are no terms" in plain text. I think hiding the sections if they are empty would be great.

nikolas: I've updated these to be consistent, in plain text.

PR #3071, opened 2020-09-04
hide btn when screen is small

ndittren:


sdreher:
Nice. Very simple solution and works for all use cases.

PR #3089, opened 2020-09-10
course settings menu 

ndittren:
Implement new design for course settings side nav bar. Medt-3624

### DIFF
--- a/mediathread/templates/dashboard/base_course_dashboard.html
+++ b/mediathread/templates/dashboard/base_course_dashboard.html
@@ -37,7 +37,7 @@
             <nav aria-label="context" role="navigation">
                 <ul class="section-tabs">
                     <li>
-                        <b>Preferences</b>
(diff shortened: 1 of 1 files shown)


mediathread/templates/dashboard/base_course_dashboard.html
```diff
@@ -36,13 +36,29 @@
         <div class="layout-horizontal-left">
             <nav aria-label="context" role="navigation">
                 <ul class="section-tabs">
+                    <li>
+                        <b>Preferences</b>
```
nikolas, line 40: the `<strong>` tag is generally used instead of `<b>` since HTML5, since it's more semantic... even though they pretty much do the same thing. https://www.w3schools.com/tags/tag_b.asp

ndittren: Ah! Got it. Made the change!

PR #3093, opened 2020-09-10
Fix tag count bug - MEDT-3630

nikolas:
Just sync with the server when leaving asset detail view - it fixes this
issue and resolves potential problems with term state, asset state, etc.

### DIFF
--- a/media/js/src/CollectionTab.jsx
+++ b/media/js/src/CollectionTab.jsx
@@ -11,7 +11,7 @@ import LoadingAssets from './alerts/LoadingAssets';
 import NoAssetsFound from './alerts/NoAssetsFound';
 
 import {
-    getAsset, getAssets, getCourseUrl, updateAsset, filterSelections
+    getAsset, getAssets, getCourseUrl
 } from './utils';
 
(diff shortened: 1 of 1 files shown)


media/js/src/CollectionTab.jsx
```diff
@@ -111,20 +111,20 @@ export default class CollectionTab extends React.Component {
...
-                // Filter the annotations of this selection based on
-                // the current asset filter state.
-                filterSelections(
-                    this.state.selectedAsset,
-                    this.state.title, this.state.owner,
-                    this.state.tags, this.state.terms,
-                    this.state.date
-                ));
-            this.props.onUpdateAssets(newAssets);
+            // Sync assets and tag/term state with the server.
```
sdreher, line 114: This is working *very well*, and does not seem slow. The filter values (owner, title, etc) are correctly retained as I toggle back and forth between full collection and item details. There is a bug with page numbers. Looking at "All Class Members" in course 84, there are 4 pages of items. Clicking into an item then "Back to Collection" results in the pagination controls displaying `1 of 1`.

PR #3094, opened 2020-09-11
new assignment table redesign

ndittren:
Per Marc's design https://mrcrymnd.github.io/mediathread/assignments-table.html


sdreher:
This looks good. I do like the delete link rather than the big red button. And you caught the bug in the filter! Nice job.

PR #3096, opened 2020-09-11
instructor only view

ndittren:


mediathread/templates/projects/sequence.html
```diff
@@ -151,6 +151,17 @@ <h5 class="modal-title">Save Changes</h5>
                                             </label>
                                         </li>
                                         {% endif %}
+                                        <li>
```
sdreher, line 154: This looks good. One minor fix - Can you move the "Instructor - only author(s) and instructor can view" option after the "Draft - only you can view"? That's the order we use in other assignments.

ndittren: Got it!

PR #3098, opened 2020-09-11
accessibility changes for collection and item space

ndittren:
*will need a new bundle.js

### DIFF
--- a/media/js/src/filters/Filter.jsx
+++ b/media/js/src/filters/Filter.jsx
@@ -353,7 +353,7 @@ export default class Filter extends React.Component {
                             </div>
                         </div>
                         <div className="form-group col-md-2">
-                            <label htmlFor="owner-filter">Owner</label>
+                            <label htmlFor="react-select-2-input">Owner</label>
                             <Select
                                 id="owner-filter"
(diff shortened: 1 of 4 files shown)


media/js/src/filters/Filter.jsx
```diff
@@ -353,7 +353,7 @@ export default class Filter extends React.Component {
                             </div>
                         </div>
                         <div className="form-group col-md-2">
-                            <label htmlFor="owner-filter">Owner</label>
+                            <label htmlFor="react-select-2-input">Owner</label>
```
nikolas, line 356: Do you need to change the Select id's here too, below this?

ndittren: Hi @nikolas! Um, I mean I don't think so? The IDs that are in the Select components are used for the encompassing DIV on the dom. And then nested several divs in is an Input tag which is where I am getting the IDs that the Label should reference (for accessibility purposes). I think the Select ids can stay as they are if Im not mistaken.

nikolas: Ah okay - i guess react-select just isn't behaving like i'd expect it to.

PR #3099, opened 2020-09-11
Add asset Rename feature - MEDT-3478

nikolas:
Screenshot_2020-09-11_15-53-10

I'm having trouble with the API unit test for staff users, so that's commented out at the moment. I just wanted to open this pull request now because it is in fact in a working state, and I've been working on this for a bit. All the pieces are in place now.

### DIFF
--- a/mediathread/assetmgr/tests/test_apiviews.py
+++ b/mediathread/assetmgr/tests/test_apiviews.py
@@ -1,0 +1,80 @@
+from rest_framework import status
+from rest_framework.test import APITestCase
+from django.urls import reverse
+
+from mediathread.factories import (
+    AssetFactory, UserFactory, MediathreadTestMixin
+)
+from mediathread.assetmgr.models import Asset
+from mediathread.sequence.tests.mixins import LoggedInTestMixin
+
+
+class AssetUpdateTestsAnon(MediathreadTestMixin, APITestCase):
+    def test_create_sherdnote_as_anon(self):
+        """
+        Ensure we can't update this asset.
+        """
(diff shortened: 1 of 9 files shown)


nikolas:
Oh, this probably also needs validation when trying to rename the asset to an empty string. That can be added before or after this PR is merged... it's up to the reviewer.

mediathread/assetmgr/tests/test_apiviews.py
```diff
@@ -0,0 +1,109 @@
...
+        }, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        asset.refresh_from_db()
+        self.assertEqual(Asset.objects.count(), 1)
+        self.assertEqual(asset.title, 'New Title')
+
+
+# TODO:
```
sdreher, line 83: Maybe we should just remove the ability for staff to rename? It's an odd corner case. If a staff needs to rename an asset, they could easily accomplish this in the backend or by impersonation.

sdreher (commented):
This looks good overall. I'm so happy to have you're starting to develop new REST apis rather than continuing on with TastyPie.

Three comments:
* Voting for adding the empty string validation in this PR, just for completeness.
* When I rename an item then navigate back to the collection, the old item name appears rather than the new name.
* Likely we can just remove the staff-level functionality here. There's no need to codify that flow when there are other options.

nikolas:
@sdreher I updated this branch and added a new commit. Staff permission for rename is removed, and validation on empty title is added.

PR #3102, opened 2020-09-12
Display the full image in the detail view - MEDT-3595

sdreher:
This should resolve a bug where the image thumbnail was displayed in the detail view rather than the full image.


nikolas:
Good fix - i'm suprised, and not sure why this was set to thumbnail!

PR #3114, opened 2020-09-15
Allow tags to contain spaces - MEDT-3767

nikolas:
My confusion here was due to some misleading documentation in
django-tagging:
https://django-tagging.readthedocs.io/en/develop/#field-types

This states that TagField is a space-separated CharField. In fact, it's
comma-separated:

    (Pdb) p note.tags
    'my tag,green,black'

Either this documentation is out of date, or Mediathread is doing
something special with tags.

### DIFF
--- a/media/js/src/forms/EditSelectionForm.jsx
+++ b/media/js/src/forms/EditSelectionForm.jsx
@@ -18,17 +18,13 @@ export default class EditSelectionForm extends SelectionForm {
         // components.
         const rawTags = this.tagsRef.current.state.value;
 
-        // Tags are handled as a space-separated CharField, while
+        // Tags are handled as a comma-separated CharField, while
         // Terms are handled with primary keys.
         let tags = '';
         if (rawTags) {
             rawTags.forEach(function(tag) {
-                tags += tag.value + ' ';
+                tags += tag.value + ',';
             });
-        }
-
-        if (tags.length) {
-            tags = tags.slice(0, -1);
         }
 
         let terms = null;
--- a/media/js/src/assetDetail/CreateSelection.jsx
+++ b/media/js/src/assetDetail/CreateSelection.jsx
@@ -17,17 +17,13 @@ export default class CreateSelection extends SelectionForm {
         // components.
         const rawTags = this.tagsRef.current.state.value;
 
-        // Tags are handled as a space-separated CharField, while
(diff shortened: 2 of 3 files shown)


media/js/src/forms/EditSelectionForm.jsx
```diff
@@ -18,12 +18,12 @@ export default class EditSelectionForm extends SelectionForm {
...
         const rawTags = this.tagsRef.current.state.value;
 
-        // Tags are handled as a space-separated CharField, while
+        // Tags are handled as a comma-separated CharField, while
         // Terms are handled with primary keys.
         let tags = '';
         if (rawTags) {
             rawTags.forEach(function(tag) {
-                tags += tag.value + ' ';
+                tags += tag.value + ',';
```
sdreher, line 26: For the example I was testing with: `a b c`, this worked correctly by appending a comma at the end of the string like so `a b c,`. However, at Ln. 30, the last character is removed from the string, in this case the comma. The tags then separate again on save, resulting in `a`, `b`, `c`. I *think* the code from Ln. 30-32 can simply be removed, but wanted you to review first.

nikolas: Tags that contain commas in their names, like `a b c,` will be changed to `a b c` because it's saved as a comma-separated string.

The last comma is always removed from this string at line 31, because it's not necessary. That code removes the last comma that was added by the code at line 26.

I'm trying to reproduce the issue here but I can't see it... What string do I save to get it to get it to split into three separate `a`, `b`, `c` tags?

sdreher: To reproduce:
* Find a selection with no tags.
* Type in `a b c`
* Press enter
* Press Save
* Open the item
* The tag is now `a`,`b`,`c`

The logic works for some reason if tags already exist, or if you're entering multiple tags at once. It does not work with a single tag with spaces.

nikolas: Ah okay - yes, removing lines 30-32 fixes the problem, like you mentioned. This is just the annoying behavior of django-tagging:

"If the input doesn’t contain any commas or double quotes, it is simply treated as a space-delimited list of tag names."

- https://django-tagging.readthedocs.io/en/develop/#tag-input

I don't think it should work that way, in my opinion! But, we can get around it by just always including a comma, even for a single tag.

PR #3155, opened 2020-09-28
Fix subquery that may return multiple rows

sdreher:
This resolves a corner case where an annotation is attempted on
an assignment where a user is both a response author and separately
a response participant.


sdreher:
Self-review. Merging to resolve a production issue.

PR #3161, opened 2020-09-30
sequence collection space switched to new style

ndittren:
* This address the collection modal for the Sequence Assignment/Sequence project spaces so that it is consistent with the new look (ex: the collection space in composition assignment).
* This also address some changes to the cypress test.
*NOTE: The pagination on the modal was causing some issues. I will open up a different branch/pr with those changes.

### DIFF
--- a/mediathread/templates/clientside/collectionwidget_assets.mustache
+++ b/mediathread/templates/clientside/collectionwidget_assets.mustache
@@ -90,19 +90,6 @@
                             -Untitled-
                         {{/title}}
                     </a>
-                    {{#citable}}
-                        <span class="citationTemplate">
-                            <img class="materialCitation clickableCitation" alt=""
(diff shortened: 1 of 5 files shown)


mediathread/templates/clientside/collectionwidget_assets.mustache
```diff
@@ -43,196 +42,153 @@
...
+                        target="_blank" href="{{local_url}}">
+                        {{#title}}
+                            {{#ellipsis35}}{{title}}{{/ellipsis35}}
+                        {{/title}}
+                        {{^title}}
+                            -Untitled-
+                        {{/title}}
+                    </a>
+                    {{#citable}}
+                        <span class="citationTemplate">
```
sdreher, line 94: This item-level citation insert can be removed.

PR #3181, opened 2020-10-06
new js bundle

ndittren:
* new js bundle for the accordion bug fix


nikolas:
This bundle isn't minified, so I'm just going to rebuild this myself and make a new push to master. It should be rebuilt with the command `npm run build`.

And of course.. this should really be handled automatically, by jenkins. I haven't been able to get that working, though.

PR #3187, opened 2020-10-08
pagination added to the mini collection space

ndittren:
* Added pagination space to the collection space used in Sequence projects and assignments

### DIFF
--- a/mediathread/templates/clientside/collectionwidget.mustache
+++ b/mediathread/templates/clientside/collectionwidget.mustache
@@ -2,13 +2,6 @@
 <script type="text/x-template" id="collectionwidget-template">
 
 <div class="collection_table collection-widget">
-    <div class="ajaxloader">
-        <div class="message">
-        <img alt="Loading..." src="{{#staticUrl}}{{/staticUrl}}img/ajax-loader.gif">
-        <br /><br />
-        <div>Refreshing the collection</div>
-        </div>
-    </div>
     <div class="media-column">
         <div class="filter-widget bg-light" style="display: none">
             <div class="row px-3">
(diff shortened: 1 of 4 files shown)


mediathread/templates/clientside/collectionwidget.mustache
```diff
@@ -116,6 +109,31 @@
     <div class="collection-assets annotation-embedding">
         <div id="asset_table" class="asset-table"></div>
     </div>
+
+    {{#paginator}}
```
sdreher, line 113: A slight nit -- maybe wrap the `paginator` block to only display if assets exist. You can use:
```
   {{#asset_count}}
      ...all the paginator code
   {{/#asset_count}}
```

mediathread/templates/clientside/collectionwidget.mustache
```diff
@@ -116,6 +109,31 @@
     <div class="collection-assets annotation-embedding">
         <div id="asset_table" class="asset-table"></div>
```
sdreher, line 110: The old loader needs to be replaced with the one from the composition space. Right above the `asset_table` here, you can add in:
```
        <div class="ajaxloader">
            <div class="message">
            <img alt="Loading..." src="{{#staticUrl}}{{/staticUrl}}img/ajax-loader.gif">
            <br /><br />
            <div>Refreshing the collection</div>
            </div>
        </div>
```

mediathread/templates/clientside/collectionwidget.mustache
ndittren: Got it!

mediathread/templates/clientside/collectionwidget.mustache
ndittren: @sdreher when I had the ajaxloader code in the collectionwidget.mustache, it would display the loader but only to about half way down. I placed the loader in the modal dialog div so that the loader styles would apply to the whole modal. I can place it back in here but the styling would have to change a little.

PR #3219, opened 2020-10-17
Revert the webpack 5.0 and webpack-cli updates

sdreher:
This upgrade broke the webpack dev flow. We'll need to take a look at how Nick handled the upgraded in Locus Tempus and make the same changes here.


nikolas:
This is solved here: #3226

PR #3248, opened 2020-10-26
Student response widget

ndittren:
* changed the response dropdown for the instructor to reflect new design

### DIFF
--- a/mediathread/templates/projects/composition_assignment.html
+++ b/mediathread/templates/projects/composition_assignment.html
@@ -120,15 +120,21 @@
                         <div class="dropdown d-inline ml-1 response-dropdown">
                             <button class="btn btn-primary dropdown-toggle" type="button" id="student-response-dropdown"
                                 data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                                {{responses|length}} of {{students|length}} students responded
+                                Student Responses
                             </button>
                             <div class="dropdown-menu dropdown-menu-right" aria-labelledby="student-response-dropdown">
                                 {% for student in students %}
                                     {% student_response responses student as response %}
-                                    {% if response %}
+                                    {% feedback responses student as feedback %}
+                                    {% if response.date_submitted %}
                                         <a href="{% url 'project-workspace' request.course.id response.id %}"
                                             class="dropdown-item" type="button">
-                                            {% public_name for student %} ({{response.date_submitted|date:"m/d/Y"}})
+                                            {% public_name for student %}
+                                            {% if feedback %}
+                                                <span class="badge badge-success">Complete</span>
+                                            {% else %}
+                                                <span class="badge badge-warning">Respond</span>
+                                            {% endif %}
                                         </a>
                                     {% else %}
                                         <a href="#" class="dropdown-item disabled" type="button">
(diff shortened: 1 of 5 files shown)


mediathread/templates/projects/composition_assignment.html
```diff
@@ -120,22 +120,39 @@ <h1 class="page-title">
...
-                                            {% public_name for student %} ({{response.date_submitted|date:"m/d/Y"}})
-                                        </a>
-                                    {% else %}
-                                        <a href="#" class="dropdown-item disabled" type="button">
-                                            {% public_name for student %}
-                                        </a>
-                                    {% endif %}
-                                {% endfor %}
+                                <h6 class="dropdown-header">Submitted
+                                    {% for student in students %}
```
sdreher, line 128: Rather than iterating students here...you might be able to iterate `submitted`. This could make the page a bit peppier as getting a student's response & feedback can be heavy operations.

mediathread/templates/projects/composition_assignment.html
```diff
@@ -120,22 +120,39 @@ <h1 class="page-title">
...
+                                        {% student_response responses student as response %}
+                                        {% feedback responses student as feedback %}
```
sdreher, line 130: I suggest moving the `feedback` retrieval within the `if response.data_submitted` clause to minimize the data retrieval here.

mediathread/templates/projects/sequence_assignment_responses.html
```diff
@@ -19,22 +19,37 @@
...
-                        {% public_name for student %} ({{response.date_submitted|date:"m/d/Y"}})
-                    </a>
-                {% else %}
-                    <a href="#" class="dropdown-item disabled" type="button">
-                        {% public_name for student %}
-                    </a>
-                {% endif %}
-            {% endfor %}
+            <h6 class="dropdown-header">Submitted
+                {% for student in students %}
```
sdreher, line 27: Same thoughts here on `submitted` vs iterating all the students.

mediathread/templates/projects/composition_assignment.html
```diff
@@ -120,22 +120,39 @@ <h1 class="page-title">
...
+                                        {% if response.date_submitted %}
+
+                                                <a href="{% url 'project-workspace' request.course.id response.id %}"
+                                                    class="dropdown-item" type="button">
+                                                    {% public_name for student %}
```
sdreher, line 135: The student's name is coming through in all capitals. I like the header in all caps, but find it difficult to read the names. Can you return them to camel case?

mediathread/templates/projects/composition_assignment.html
ndittren: @sdreher right! Okay I refactored a bit so it's not so heavy.

mediathread/templates/projects/composition_assignment.html
```diff
@@ -120,22 +120,32 @@ <h1 class="page-title">
...
-                                    {% student_response responses student as response %}
-                                    {% if response.date_submitted %}
-                                        <a href="{% url 'project-workspace' request.course.id response.id %}"
+                                <h6 class="dropdown-header">Submitted</h6>
+                                {% for submitted in submittedResponses %}
+                                    {% feedback submitted as feedback %}
+                                        <a href="{% url 'project-workspace' request.course.id submitted.id %}"
                                             class="dropdown-item" type="button">
-                                            {% public_name for student %} ({{response.date_submitted|date:"m/d/Y"}})
+                                            {{submitted.attribution}}
```
sdreher, line 132: Any reason not to use "public_name" here as you do in the loop below?

ndittren: I think you can only derive public_name from "students" or user data and since I was iterating through responses I had to use attribution to get the authors.

sdreher: Ah okay, I get it.

Great work on this. I'll merge now!

PR #3285, opened 2020-11-11
Refactor pagination to be independent react component

nikolas:
Natalia and I have tested this out, but we need to do some more QA work - feedback welcome!

I'm having some trouble QA'ing this myself because each getAssets request for "All Students" takes like 45 seconds! I even have memcached set up on my dev system... I'd like to get to the bottom of this issue. So, it may be much easier to debug on staging, which seems like a fine approach.


ndittren:
I did some additional QA-ing locally since the requests are slightly faster on my machine (averaging around 18 seconds).  Things look like they are working correctly. Ran `eslint` and `make test` as well as the `cypress` tests. All pass.

PR #3296, opened 2020-11-18
new thumbnail images for when media is missing

ndittren:
* This PR adds in the new icons for missing thumbnails in the collection space.
* This also addresses a small bug where the badges weren't displaying with the standard [image, video, audio].
* I added in a 4th type with 'unknown'. I'm not too sure if that is the correct language. 
* Requires a new bundle.js.


media/js/src/Asset.js
```diff
@@ -31,13 +28,12 @@ export default class Asset {
             const url = `https://vimeo.com/api/oembed.json?url=${vUrl}`;
             return fetch(url).then(function(response) {
                 if (response.status === 200) {
-                    return response.json();
+                    return response.json().then(function(d) {
```
sdreher, line 31: This works nicely. I'm not sure a `then` function is needed...could this be simplified?

media/js/src/utils.js
```diff
@@ -326,6 +326,10 @@ const getAssetType = function(primaryType) {
         type = 'video';
     } else if (primaryType === 'image_fpxid' || primaryType === 'image_fpx') {
         type = 'image';
+    } else if (primaryType === 'mp4_audio') {
+        type = 'audio';
+    } else {
+        type = 'unknown';
```
sdreher, line 332: `primary_type` can also === 'image' here. For reference, all the labels are listed here: https://github.com/ccnmtl/mediathread/blob/master/mediathread/assetmgr/models.py#L154

PR #3335, opened 2020-12-01
Mediaprefix add

ndittren:
* added media prefix so the thumbnails show up in production

### DIFF
--- a/media/js/src/GridAsset.jsx
+++ b/media/js/src/GridAsset.jsx
@@ -200,13 +200,15 @@ export default class GridAsset extends React.Component {
                 })
             });
         }
+        const mediaPrefix = typeof MediaThread !== 'undefined' ?
+            window.MediaThread.staticUrl : '/media/';
 
         if(this.asset.getType() === 'unknown'){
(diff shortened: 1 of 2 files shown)


media/js/src/GridAsset.jsx
```diff
@@ -200,13 +200,15 @@ export default class GridAsset extends React.Component {
...
+            window.MediaThread.staticUrl : '/media/';
 
         if(this.asset.getType() === 'unknown'){
-            this.setState({thumbnailUrl: '/media/img/thumb_unknown.png'});
+            this.setState({thumbnailUrl: mediaPrefix + 'img/thumb_unknown.png'});
         }
 
         if(this.asset.getType() === 'audio'){
-            this.setState({thumbnailUrl: '/media/img/thumb_audio.png'});
+            this.setState({thumbnailUrl: mediaPrefix + '/media/img/thumb_audio.png'});
```
nikolas, line 211: Does the `/media/` at the beginning of this string need to be taken out?

ndittren: Ah! Yes thank you. Just fixed!

PR #3361, opened 2020-12-11
Implementation of the clear all button for the main collection space

ndittren:
* Clear all button added to the main collection space with the other filters

### DIFF
--- a/media/js/src/CollectionTab.jsx
+++ b/media/js/src/CollectionTab.jsx
@@ -65,7 +65,8 @@ export default class CollectionTab extends React.Component {
                 'all',
             tags: [],
             terms: [],
-            date: 'all'
+            date: 'all',
+            currentPage: 0
         });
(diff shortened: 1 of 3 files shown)


media/js/src/CollectionTab.jsx
```diff
@@ -56,6 +57,18 @@ export default class CollectionTab extends React.Component {
         });
     }
 
+    onClearFilter() {
+        this.setState({
```
sdreher, line 61: This is working correctly for page 1. If I attempt to "Clear" on page > 1 then things get a bit confused.

ndittren:
@sdreher There's some wonky things happening with the button when the screen is smaller. We could fix it with css or maybe just have it say 'Clear'?

sdreher:
@ndittren - Let's go with just "Clear". Good suggestion.

PR #3363, opened 2020-12-11
New promo video added to the LTI landing page

ndittren:
* New promo video added to the LTI landing page

### DIFF
--- a/mediathread/templates/lti_auth/landing_page.html
+++ b/mediathread/templates/lti_auth/landing_page.html
@@ -45,6 +45,7 @@
                 <p class="lead text-center">
                     <a class="btn btn-primary btn-lg" href="{{landing_url}}" target="_blank">Launch Mediathread</a>
                 </p>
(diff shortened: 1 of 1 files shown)


mediathread/templates/lti_auth/landing_page.html
```diff
@@ -45,6 +45,7 @@
                 <p class="lead text-center">
                     <a class="btn btn-primary btn-lg" href="{{landing_url}}" target="_blank">Launch Mediathread</a>
                 </p>
+                <iframe width="560" height="315" src="https://www.youtube.com/embed/xwy3I1AMLH4" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
```
sdreher, line 48: Could the video be centered? I think the easiest way to accomplish this is to wrap the iframe in a div. Something like `<div class="text-center w-100">`.

PR #3399, opened 2020-12-29
keyboard pan enabled.

ndittren:
* Enables panning with arrow keys in the Item Detail space.

### DIFF
--- a/media/js/src/assetDetail/AssetDetail.jsx
+++ b/media/js/src/assetDetail/AssetDetail.jsx
@@ -905,6 +905,7 @@ export default class AssetDetail extends React.Component {
 
             this.map = new Map({
                 target: `map-${this.props.asset.id}`,
+                keyboardEventTarget: document,
                 layers: [
                     new ImageLayer({
                         source: new Static({


media/js/src/assetDetail/AssetDetail.jsx
```diff
@@ -905,6 +905,7 @@ export default class AssetDetail extends React.Component {
 
             this.map = new Map({
                 target: `map-${this.props.asset.id}`,
+                keyboardEventTarget: document,
```
sdreher, line 908: Excellent - I'm glad this was a simple change. I'm looking at it right now and don't see how I can interact using the keyboard? Can you point me in the right direction?

ndittren: @sdreher Sure! Once you're in the item detail space, you can move the image with the arrow keys. This PR also requires a new bundle.js.

sdreher: Cool! I'm able to move the image with the keys. Thanks for the instructions & great research and implementation.

Going forward, we'll probably want to refine the interaction a bit, enable/disable kind of like we're doing for the Zoom in/out keys and likely add some help text.

PR #3428, opened 2021-01-13
Center control

ndittren:
* A control added for image annotation that allows the user to recenter the image.


media/js/src/assetDetail/AssetDetail.jsx
```diff
@@ -870,6 +873,8 @@ export default class AssetDetail extends React.Component {
                 defaultControls().forEach(function(control) {
                     me.map.addControl(control);
                 });
+                me.map.addControl(new CenterControl());
```
sdreher, line 876: When testing this, the first-time through, everything works. On viewing an image, Zoom & Center are perfect. On click "draw shape", all controls disappear as expected and I'm able to draw a shape on the image. However, the next time through, on "draw shape", the center buttons disappears and the zoom buttons stay. The next time through, the zoom buttons disappear, but the center buttons stay. Let's talk about the best way to tackle this.

media/js/src/centercontrol.js
```diff
@@ -0,0 +1,31 @@
+import {Control} from 'ol/control';
+
+const CenterControl = /*@__PURE__*/(function(Control) {
```
sdreher, line 3: Nice job moving this out into its own control! :clap:

sdreher (changes requested):
Let's take a quick look at this together. There's some quirkiness in the show/hide controls.

sdreher (approved):

PR #3606, opened 2021-03-16
Student assignment list view: Fix inconsistent response filter

sdreher:
A discussion assignment "response" is modeled quite differently than a composition, sequence or selection response. The simple filter query being completed server side for "no response yet", doesn't work for discussion assignment. To work around this, implement a simple client-side filter that checks for status and comment count.

### DIFF
--- a/mediathread/projects/templatetags/user_projects.py
+++ b/mediathread/projects/templatetags/user_projects.py
@@ -104,5 +104,5 @@ def date_format_change(date):
 
 
 @register.simple_tag
-def show_discussion_response(status, comments):
-    return status != 'no-response' or comments[0] == 0
+def show_discussion_response(status, comment_count):
+    return status != 'no-response' or comment_count == 0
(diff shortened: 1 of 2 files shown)


mediathread/projects/templatetags/user_projects.py
```diff
@@ -101,3 +101,8 @@ def date_format_change(date):
     newFormat = ''.join(newDate.split())
 
     return re.sub(r'(?<=[,])(?=[^\s])', r' ', newFormat)
+
+
+@register.simple_tag
+def show_discussion_response(status, comments):
```
nikolas, line 107: It's not clear to me from the code what the `comments` list is. Is this a list of comments? Comment lengths? A note would be good to clarify this.

sdreher: Good point. I've updated to clarify.

PR #3648, opened 2021-03-29
Line string annotation tool

ndittren:
* A different version of the freeform tool that doesn't close the annotation or have a fill.


sdreher:
This is looking okay...but the cancel/clear/release buttons seem really confusing to me. Also, the free-form line should render more brightly so that it is easier to see.

PR #3661, opened 2021-03-31
Annotation tools and workflow

ndittren:
* Two tools added to the annotation space as well as a change in the overall create selection flow as noted in our user stories.


sdreher (commented):
These changes are looking really good. I am seeing a few weirdnesses in behavior. Listing them here. I'm also happy to demonstrate if these descriptions aren't clear.
* On editing an annotation and canceling, the original annotation disappears. A page refresh shows the annotation is still there. Looks like there needs to be some logic to redraw the original annotation.
* On editing an annotation, sometimes there are two Selection labels
* On creating or editing, use the standard flow to draw a shape. The workspace returns correctly to the "View" state, with Clear button and Zoom/Pan buttons enabled. Then, click another button -- both Clear & Cancel are now displayed.

sdreher:
Nice - this is looking much smoother. Reporting a few more bugs:
* Creating a "line squiggle" works perfectly. However, when I then choose to view a different selection, the line squiggle remains on the image.
* When I choose to edit an existing selection, should the clear button be there?
* In general...the flow is a bit wonky around edit, I can demonstrate today when we meet.

PR #3727, opened 2021-04-29
Shows default image when no thumbnail is available

ndittren:
* The default image thumbnail sometimes wouldn't show up and would cause an error. This makes sure the url is updated to the default image thumbnail in the case where a thumbnail cannot be produced.

### DIFF
--- a/media/js/src/Asset.js
+++ b/media/js/src/Asset.js
@@ -59,6 +59,9 @@ export default class Asset {
         if (!this.asset.sources.thumb){
             this.asset.sources.thumb = mediaPrefix + 'img/thumb_image.png';
         }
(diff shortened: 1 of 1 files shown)


media/js/src/Asset.js
```diff
@@ -58,6 +58,7 @@ export default class Asset {
             window.MediaThread.staticUrl : '/media/';
         if (!this.asset.sources.thumb){
             this.asset.sources.thumb = mediaPrefix + 'img/thumb_image.png';
+            this.asset.sources.image.url = mediaPrefix + 'img/thumb_image.png';
```
sdreher, line 61: In certain situations...this wipes out the image and thumbnail display.

PR #3744, opened 2021-05-04
Update CORS setting name

nikolas:
The `CORS_ORIGIN_ALLOW_ALL` setting is now called
`CORS_ALLOW_ALL_ORIGINS`.

https://pypi.org/project/django-cors-headers/


mediathread/settings_shared.py
```diff
@@ -179,7 +179,7 @@ def no_reject(request, reason):
 
 ACCOUNT_ACTIVATION_DAYS = 7
 
-CORS_ORIGIN_ALLOW_ALL = True
+CORS_ALLOW_ALL_ORIGINS = True
```
sdreher, line 182: lol - was that really necessary?

sdreher, line 182: lol - was that really necessary?

nikolas: hah i know right. So, the old setting name is still working - this is just futureproofing it. This doesn't fix the Dartmouth error - I'm still looking into that, able to reproduce it.

PR #3902, opened 2021-07-20
Improve PDF annotation behavior - MEDT-3943

nikolas:
* Only allow one rectangle per annotation
* Add annotation toolbar basics

### DIFF
--- a/media/js/pdfjs/AnnotationController.js
+++ b/media/js/pdfjs/AnnotationController.js
@@ -36,50 +36,80 @@ const getCanvasCoords = function(x1, y1, x2, y2) {
  */
 class AnnotationController {
     constructor() {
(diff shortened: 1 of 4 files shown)


sdreher:
The code looks fine to me. I'm not able to verify locally as I seem to be running into that race condition where the state does not change into "annotation mode". Any tricks to make that happen?

nikolas:
Ah okay - yeah, that sounds like the bug I need to address. Interesting that it's always happening on your system. I can reproduce it about half the time but I don't have reproducible steps to do this. I'll prioritize that bug and see if I can fix that this afternoon.

sdreher:
No worries. I'm happy to merge this as is. I'll do some more comprehensive testing when the other bug is fixed.

PR #3922, opened 2021-07-27
Save pdfjs annotation to back-end - MEDT-3938

nikolas:
* Fix page number bug
* Create pdf annotation on save

### DIFF
--- a/media/js/pdf/AnnotationController.js
+++ b/media/js/pdf/AnnotationController.js
@@ -124,6 +124,12 @@ class AnnotationController {
 
     closeRect(x, y) {
         this.rect.coords.push([x, y]);
+
+        // Tell the parent application about the new rectangle.
+        window.top.postMessage({
+            message: 'pdfAnnotationRectCreated',
+            rect: this.rect
+        }, '*');
     }
 
     displayRect(x1, y1, x2, y2, scale=1) {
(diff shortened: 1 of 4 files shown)


mediathread/djangosherd/models.py
```diff
@@ -357,6 +357,14 @@ def __str__(self):
                                              self.asset.title,
                                              self.asset.course.title)
 
+    def is_global_annotation(self):
+        # Disable "global annotation" stuff for PDF type. Let's phase
```
sdreher, line 361: I would *so enjoy* removing global annotations from the codebase.

mediathread/assetmgr/models.py
```diff
@@ -270,6 +270,12 @@ def global_annotation(self, user, auto_create=True):
 
     def media_type(self):
         label = 'video'
+
+        try:
```
sdreher, line 274: Is there a situation where primary does not exist?

nikolas: During testing, with the new is_global_annotation() change, some of the reports tests fail with Source.DoesNotExist. Debugging this was pretty confusing - we can possibly solve this another way, or postpone the global annotation changes altogether by just using the hardcoded ranges in the new PDF annotation structure.

sdreher: I like your suggestion of just using hard-coded values in the ranges. To me, that feels cleaner then adding additional exception code. What's your feeling?

nikolas: Sure, that's fine - i mean that's what we're currently doing for image annotations. That's what I was trying to get rid of in the first place, but I can just do that for now if we don't want to spend time re-thinking global annotations yet.

sdreher: Yeah - removing one workaround for images/pdf seems simpler than removing two. That said, I'm fine talking through how we can remove global annotations now. No reason to put that off.

PR #3951, opened 2021-08-06
Enable pdf annotation editing

nikolas:


sdreher (commented):
Reviewing this in my local dev env, the selection edit form appears as expected. But, I'm not seeing the Rectangle tool appear. The form also will not save, the exception is:
`bundle.js:2 Uncaught TypeError: Cannot read property 'coords' of null`

What's the expectation for this PR -- is it just prepping for the edit functionality?

nikolas:
This change allows for a pdf annotation to have a new rectangle selected and saved via the Edit dialog. Apparently there are errors, though - I'm taking a look. I only tested once or twice and it worked for me so far.

nikolas:
The error you're seeing indicates that you're not building the new JS here - run `npm run dev` alongside make runserver. You should then see the rectangle tool appear in the edit view, and that coords error fixed.

PR #3980, opened 2021-08-27
Add pdf selections to collection space

nikolas:
This is implemented on a custom SVG layer and drawn with SVG.js.

There is a minor placement/scaling issue with the annotation positioning here, but the basics are in place.

Screenshot_2021-08-27_16-14-36


sdreher (commented):
This is looking good as a first step! I am seeing issues with placement, and page 2 annotations were not working. I think this can merge as is though. Nice progress.

PR #4023, opened 2021-09-16
Fix pdf annotation positioning on collection - METH-3910

nikolas:
* Scroll to correct page number when flipping through annotations
* Fix annotation position as window is resized.
    
There are more issues I've seen with certain a more complex PDF I'm
testing with - a 300 page book where some of the pages are different
aspect ratios. I'll be handling those bugs separately


sdreher (commented):
The positioning of selections on page 1 of a PDF looks more accurate.

I am seeing an odd behavior where clicking once to the right does not show the first selection.

sdreher (commented):
The behavior is looking very good in my dev env. I'm still seeing the weird "too tall" issue, but the page navigation and selection positioning looks good.

PR #4035, opened 2021-09-20
Scroll to correct annotation page number on pdf detail view

nikolas:


sdreher (commented):
Looks good in my dev env. 🎉

PR #4187, opened 2021-11-01
PDF annotation improvements in sherd.js

nikolas:
* Constrain pdf view on width as well as height.
* Add an annotation-aware mode for renderPage(). Currently this reads
  the annotation geometry and does some scaling based on that, putting
  the annotation at roughly the top-left portion of the view. Next steps
  here are to perfect this and center it.


sdreher (commented):
Better and better. My one thought is that maybe top center would be a better position? That can keep until you've finished iterating on this feature.

Oh, and the bug I reported before where the thumbnail overlays buttons & other links does still exist. :-(

PR #4191, opened 2021-11-03
Change Annotation.is_global_annotation to a boolean field

nikolas:
Don't base this on the presence of range1 and range2 - that causes problems with other parts of the code, namely the mustache timecode display issues.

I know we've discussed removing global annotations altogether - obviously that would be a big change that touches a lot of the SherdNote code. I think these changes here at least make things a bit clearer and more explicit to make way for those changes in the future.


mediathread/djangosherd/models.py
```diff
@@ -20,14 +20,14 @@
 from structuredcollaboration.models import Collaboration
 
 
-NULL_FIELDS = dict((i, None) for i in
-                   'range1 range2 title'.split())
+NULL_FIELDS = {'title': None}
```
sdreher, line 23: NULL_FIELDS is referenced in `djangosherd`'s `create_annotation`. (https://github.com/ccnmtl/mediathread/blob/master/mediathread/djangosherd/views.py#L19)

I *think* the `create_annotation` code is making use of NULL_FIELDS to ensure the user is specifying a portion of an asset rather than the "whole" asset. But likely worth tracing through to determine exact behavior.

mediathread/djangosherd/models.py
```diff
@@ -216,16 +211,24 @@ def global_annotation(self, asset, author, auto_create=True):
...
+            gannotation, created = self.get_or_create(**args)
         else:
             try:
                 gannotation = self.get(**args)
             except SherdNote.DoesNotExist:
                 gannotation = None
+
             return gannotation, False
 
+        if gannotation and not gannotation.is_global_annotation:
```
sdreher, line 226: I think setting the flag here might be unnecessary if you add the `is_global_annotation` to the args above, e.g. `args.update(is_global_annotation=True)`. This also might prevent some oddness in the `get_or_create` at Ln. 217. The `get_or_create` used to retrieve a unique annotation for an asset/user combination by  checking for an empty title and empty range. As written now, it's retrieving a unique annotation for an asset/user combination simply with an empty title. That might not be quite as unique.

sdreher (commented):
Excellent refactoring exercise here. I've pointed out two areas that I think should be reviewed and possibly addressed before a merge. Once that happens, I'll pull it down locally and complete another round of testing.

mediathread/djangosherd/models.py
nikolas: Yeah, I noticed that. I removed range1 and range2 here in order to make some tests pass and just clean things up, because this NULL_FIELDS is used in this file as well.

I've added a commit that uses the original version of NULL_FIELDS in djangosherd/models.py to account for this cliipping thing.. let me know if that doesn't make sense or you think of a better way to go about this.

nikolas: I'm also just going to remove NULL_FIELDS from this models.py file, as it's really not necessary.

mediathread/djangosherd/models.py
nikolas: Yeah, I think you're right. I've made that change in a new commit here.

sdreher (commented):
Things are looking good from a code perspective. And, I've run through a testing round locally. Existing collections are intact. Uploading an image properly adds the item to the user's collection. I'm going to merge and do another round of testing in staging.

PR #4204, opened 2021-11-09
Fix bug where scale is applied twice + more pdf annotation improvements

nikolas:
Fix the width constraint issue and some surrounding issues.

I've also included some general improvements and basic centering to the pdf annotation view.


sdreher (commented):
Excellent - the annotations are displaying perfectly in the edit view popup. :clap: I am still seeing some weirdness in the "preview" view -- only half the annotation is displaying. Merging this so you can move forward.

PR #4210, opened 2021-11-11
Docker Changes

whyrg:
* Updated Dockerfile to python3 deps (probably should move this into [django.base](https://github.com/ccnmtl/docker/blob/master/django.base) instead)
* Downgraded django-celery-results to compatible version
* (Temporarily) Removed failing test on container build.
```  
File "/app/mediathread/settings_docker.py", line 8, in <module>
    DB_NAME = os.environ.get('DB_NAME', app)
NameError: name 'app' is not defined
```
* Above appears to reference the APP ENV for default value. Instead of relying on yet another place (Dockerfile) to define a default name, default to "mediathread" right in settings_docker.py.


whyrg:
premature. Closing for now.

whyrg:
Re-opening. Thanks to @sdreher for sanity checking the above `app` env changes.

docker-compose.yml
```diff
@@ -1,9 +1,19 @@
...
   image: ccnmtl/mediathread
   environment:
-    - SETTINGS=settings_compose
+    - SETTINGS=settings_docker
+    - DB_HOST=db
+    - DB_USER=postgres
+    - DB_NAME=postgres
+    - DB_PORT=5432
+    - SECRET_KEY=CHANGEME
+    - EMAIL_BACKEND=django.core.mail.backends.dummy.EmailBackend
```
sdreher, line 14: Quick note on this setting - We should walk through how will email work in the production environment.

sdreher (commented):
I just added one comment highlighting the email settings. There's no reason this needs to block this PR -- I'll proceed with the merge.

PR #4327, opened 2022-02-01
Fix pdf annotation rendering in GridAsset view - METH-3983

nikolas:


sdreher (commented):
The blue rectangles are now appearing, and the page number is correct. However, the outline is skewed to the top/left. I can show you in my dev env.

nikolas:
@sdreher the annotation positioning bug should be addressed now - there was a scaling operation happening twice instead of once. But let me know if you're still seeing weirdness here.

sdreher:
Nice - removing the double scaling fixed the issue. Merging.

PR #4341, opened 2022-02-03
Add validation error with making pdf selection with no annotation

nikolas:


media/js/src/assetDetail/AssetDetail.jsx
```diff
@@ -280,6 +280,12 @@ export default class AssetDetail extends React.Component {
                 end: this.state.selectionEndTime
             };
         } else if (this.type === 'pdf') {
+            if (!this.state.pdfRect) {
+                this.onShowValidationError(
+                    'Please make a selection on the PDF.');
```
sdreher, line 285: Nice, the error message is appearing for me now. We'll need to do some work to make our overall messaging a bit more consistent.

PR #4349, opened 2022-02-07
PDF view: handle case where user is None

nikolas:


mediathread/assetmgr/views.py
```diff
@@ -1349,8 +1349,9 @@ class PDFViewerDetailView(LoggedInCourseMixin, DetailView):
     def dispatch(self, request, *args, **kwargs):
         r = super().dispatch(request, *args, **kwargs)
 
-        if request.user.is_superuser or \
-           request.course.is_faculty(request.user):
+        if request.user and (
```
sdreher, line 1352: Curious - with the `LoggedInCourseMixin` specified. How is it possible for user to be None? Is it that `dispatch` kicks in first?

nikolas: Good question - looking through the code path here.. 
* LoggedInCourseMixin https://github.com/ccnmtl/mediathread/blob/0fcad65d98fec497221723870eccfd48b4b3ba29/mediathread/mixins.py#L195
* django's login_required decorator: https://docs.djangoproject.com/en/2.2/_modules/django/contrib/auth/decorators/#login_required
* Course.is_member: https://github.com/ccnmtl/django_courseaffils/blob/master/courseaffils/models.py#L80

After tracing through all that I realized the answer is just in this dispatch() method here. On line 1350 we call all the above code via super().dispatch(), and store it in `r`. Then, regardless of that output we do some extra tests here at line 1352. So, it's possible for request.user to be None here, and I think this change makes sense.

sdreher: Cool, thanks for the detail. Appreciated.

PR #4364, opened 2022-02-15
Add quickedit clear/cancel for PDF

nikolas:
Screenshot_2022-02-15_15-39-32

I used some questionable jQuery tactics to reposition the toolbar above the PDF view.


sdreher (commented):
The flow looks solid in my dev env. Nice work!

PR #4368, opened 2022-02-16
PDF - show annotation in quickedit edit mode

nikolas:
I refactored some of the iframe onViewSelection code to allow for
this message to be sent and executed successfully before the PDF
is actually rendered in the iframe.


sdreher (commented):
Excellent - this is working in my dev env.

PR #4392, opened 2022-03-01
Add show/hide all selections for PDF

nikolas:


sdreher (commented):
Reviewing this in my dev env...I don't see this working in the selection assignment. Is it hooked up yet?

nikolas:
Most of the issues here have been resolved from my perspective - but still one lingering problem: when the window is resized all but one of the selections disappear.

sdreher:
@nikolas - lmk if you're ready for a review, or if you want to try to get the resize working.

nikolas:
alright, this is finally ready for another review @sdreher. Some more changes were required: saving the annotations in a state object to re-display them as the page is re-rendered on resize.

media/js/app/assetmgr/asset.js
```diff
@@ -733,6 +749,11 @@
         };
 
         this.showAnnotation = function(event, ui) {
+            if (document.forms['annotation-list-filter']) {
+                document.forms['annotation-list-filter']
+                    .elements.showall.checked = false;
```
sdreher, line 754: I'm not sure about setting the `showall` to false here. Currently, the image viewer continues to show all selections, and simply focuses on the selected one. (Image also has different colored borders, but we can leave that for another time.)

sdreher (commented):
Show all is working nicely. :tada: Please take a look at the question on setting showall to false when a single selection is clicked.

media/js/app/assetmgr/asset.js
nikolas: Alright, the showall behavior should be closer to how it works on the Image type, now.

PR #4393, opened 2022-03-02
Fix static files loading in pdf.js viewer

nikolas:


sdreher (commented):
Quick question on this -- did relative pathnames not work for the icons loaded in the css?

nikolas:
Hmmm, good question. Let me see if there's a way I can avoid using this view hack - a relative path solution may be possible since the CSS is hosted on AWS as well. I'll do some experiments on staging.

nikolas:
After more digging - it seems this should have been working as originally configured, before I changed anything. This appears to be a legitimate Safari bug: https://stackoverflow.com/q/66752308/173630

nikolas:
I'm going to upgrade to Safari 15 and see if this is fixed there.

nikolas:
@sdreher so I tested in Safari 15, and had Marc test as well (see this simple example to reproduce this bug here: https://codepen.io/alexbernotas/pen/abpbVeG)

It is in fact a Safari bug that's still around. It looks like it can be worked around by using the complete path URL - i.e. the changes here in this PR. So, I think the next step here is to merge this PR, even though it adds some complexity, and see how that works on staging in Safari.

PR #4394, opened 2022-03-02
use python 3.8 for testing

nikolas:
mediathread-staging is now on python 3.8, and prod will be soon.


sdreher (commented):
Any reason not to keep both for the time being?

nikolas:
The reason I didn't keep them both is because the integration tests have been flaky. Only running these tests once per change keeps that to a minimum.

PR #4474, opened 2022-05-09
Update same site policy

sdreher:


sdreher:
Self-review to fix the broken extension

PR #4620, opened 2022-09-15
Accessibility - labels

ndittren:
* fixes input labels 
* aria label for recenter button


media/js/src/filters/AssetFilter.jsx
```diff
@@ -252,9 +252,10 @@ export default class AssetFilter extends React.Component {
                             </div>
                         </div>
                         <div className="form-group col-md-2">
-                            <label htmlFor="react-select-2-input">Owner</label>
+                            <label htmlFor="ownerFilter">Owner</label>
                             <Select
                                 id="owner-filter"
+                                inputId='ownerFilter'
```
nikolas, line 258: html ID's must be unique on the page - so just be aware of that when implementing these sorts of things. It may be best to generate a random string or number here, just to make sure there are no conflicts with something else on the page, and do something like:
```
const uniqueId = 'abcdef'; // generate this randomly, though
const ownerFilterId = `ownerFilter-${uniqueId}`;
```
Then in the template:
```
<label htmlFor={ownerFilterId}>Owner</label>
<Select
    inputId={ownerFilterId}
```

nikolas: Honestly, all that may not be necessary though. Your convention here looks good. We can roll that out and extend on it if needed or we see any issues.

media/js/src/filters/AssetFilter.jsx
```diff
@@ -252,9 +252,10 @@ export default class AssetFilter extends React.Component {
                             </div>
                         </div>
                         <div className="form-group col-md-2">
-                            <label htmlFor="react-select-2-input">Owner</label>
+                            <label htmlFor="owner-filter">Owner</label>
```
nikolas, line 255: Also - keep in mind that messing around with the id's and `for` attributes may not even be necessary. The way I prefer to associate labels with inputs is just wrap the label around the input. This is a completely acceptable, accessible way of using the html: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

ndittren: @nikolas wrapping the label around the Select component warps the styling a bit :/

nikolas: Oh okay - yeah just doing it this way is probably easier then.

media/js/src/filters/AssetFilter.jsx
```diff
@@ -215,6 +215,12 @@ export default class AssetFilter extends React.Component {
         const tagsOptions = tagsToReactSelect(this.props.allTags);
         const termsOptions = termsToReactSelect(this.props.allTerms);
 
+        const ownerFilterId = `ownerFilter-${(Math.random() + 1).toString(36).substring(7)}`;
+        const termFilterId = `termFilter-${(Math.random() + 1).toString(36).substring(7)}`;
+        const tagFilterId = `tagFilter-${(Math.random() + 1).toString(36).substring(7)}`;
+        const dateFilterId = `dateFilter-${(Math.random() + 1).toString(36).substring(7)}`;
+        const searchFilterId = `searchFilter-${(Math.random() + 1).toString(36).substring(7)}`;
```
nikolas, line 222: This repeated `(Math.random() + 1).toString(36).substring(7)` logic can definitely be refactored out into a common function here.

media/js/src/forms/SelectionForm.jsx
```diff
@@ -173,10 +174,11 @@ export default class SelectionForm extends React.Component {
 
                 {hasVocab && (
                     <Form.Group className="mb-3">
-                        <label htmlFor="react-select-7-input">
+                        <label htmlFor="newSelectionTerms">
                             Course Vocabulary</label>
                         <Select
-                            id="newSelectionTerms"
+                            id="new-Selection-Terms"
+                            inputId='newSelectionTerms'
```
nikolas, line 181: This is minor but there is some inconsistency with the capitalization between here and the other inputs being changed in this PR. In addition, just using a different casing convention to differentiate ID's doesn't really feel semantic or meaningful - it could get confusing. I propose we stick to either dashes, or camelCase with these changes, but not both.

To differentiate between the different elements `id` and `inputId` are referring to, how about going with the convention of:
```
<Select
    id="new-selection-terms-selectcontainer"
    inputId="new-selection-terms-input"
```
This way it's self-explanatory and more obvious what's going on. `id` is really referring to react-select's container, according to the docs - https://react-select.com/props

Issue #4647, opened 2022-09-30
Implement alternative asset input modality


nikolas:
Don't rely on extension for importing assets into mediathread. Provide form input for videos/images.

nikolas:
Done-ish. Need to support Vimeo here.

nikolas:
Also TODO here:
* Better image URL input validation
* Data URLs: Should we support these with an unlimited-length TextField? I'm kind of thinking no. And if not - how do we explain this to the user?

nikolas:
Input validation done, just not fully QA'ed. 

Currently, thumbnails aren't properly set when importing external videos.

PR #4652, opened 2022-10-04
Add simple AssetCreateView  #4647

nikolas:
Create basic form for creating a Mediathread asset based on an external image or video source. Rename the new, simple `UploadedAssetCreateView` to `AssetCreateView` and use this for media import as well.

Screenshot_2022-10-03_20-25-09


ndittren:
@nikolas When I upload a video with a YouTube url, the thumbnail in the collection space doesn't show up. Instead we get the generic icon.

PR #4658, opened 2022-10-06
Update cypress github action config

nikolas:
These changes separate out the cypress run from the django tests.

* https://github.com/cypress-io/cypress/issues/22086
* https://github.com/cypress-io/github-action

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -2,6 +2,7 @@ name: build-and-test
 on: [push, workflow_dispatch]
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


nikolas:
@ndittren I think this was a Firefox v105 issue, which would make sense why it just randomly started failing the other day.

Switching back to Firefox 102.3.0esr causes the tests to run fine. The Firefox ESR releases should be a bit more stable than the main branch anyways.

The other changes in this PR break out the cypress tests to a separate github actions job from the django tests, which I think makes it a little more clear to us what's going wrong when things fail.

PR #4668, opened 2022-10-10
collection space annotation labels

ndittren:
- adding dynamic alt and aria-labels to collection space annotations

### DIFF
--- a/media/js/src/AnnotationScroller.jsx
+++ b/media/js/src/AnnotationScroller.jsx
@@ -22,15 +22,11 @@ export default class AnnotationScroller extends React.Component {
             this.setState({
                 currentAnnotation: newAnnotation
             });
-            let annotationTitle = this.props.asset.annotations[newAnnotation].metadata.title;
-            document.getElementById(`map-${this.props.asset.id}`).setAttribute(
-                'aria-label', `Annotation titled ${annotationTitle}`);
+
             this.props.onSelectedAnnotationUpdate(newAnnotation);
         } else {
             this.setState({currentAnnotation: -1});
             this.props.onSelectedAnnotationUpdate(null);
-            document.getElementById(`map-${this.props.asset.id}`).setAttribute(
-                'aria-label', `Image thumbnail for asset: ${this.props.asset.title}`);
         }
     }
     onNextClick(e) {
@@ -40,9 +36,7 @@ export default class AnnotationScroller extends React.Component {
             this.setState({
                 currentAnnotation: newAnnotation
             });
-            let annotationTitle = this.props.asset.annotations[newAnnotation].metadata.title;
-            document.getElementById(`map-${this.props.asset.id}`).setAttribute(
-                'aria-label', `Annotation titled ${annotationTitle}`);
+
             this.props.onSelectedAnnotationUpdate(newAnnotation);
         }
     }
--- a/media/js/src/AnnotationScroller.jsx
+++ b/media/js/src/AnnotationScroller.jsx
@@ -22,11 +22,15 @@ export default class AnnotationScroller extends React.Component {
             this.setState({
                 currentAnnotation: newAnnotation
             });
-
+            let annotationTitle = this.props.asset.annotations[newAnnotation].metadata.title;
+            document.getElementById(`map-${this.props.asset.id}`).setAttribute(
+                'aria-label', `Annotation titled ${annotationTitle}`);
             this.props.onSelectedAnnotationUpdate(newAnnotation);
         } else {
             this.setState({currentAnnotation: -1});
             this.props.onSelectedAnnotationUpdate(null);
+            document.getElementById(`map-${this.props.asset.id}`).setAttribute(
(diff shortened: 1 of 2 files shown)


zmustapha:
In `media/js/src/GridAsset.jsx` line 176 in this div `<div className="card-thumbnail">`, add `aria-live="polite"` attribute. VoiceOver announces the change of alternative text when you go through the thumbnails.

media/js/src/AnnotationScroller.jsx
```diff
@@ -37,7 +56,9 @@ export default class AnnotationScroller extends React.Component {
                 currentAnnotation: newAnnotation
             });
 
-            this.props.onSelectedAnnotationUpdate(newAnnotation);
```
nikolas, line 40: Removing `onSelectedAnnotationUpdate` prevents the annotation view from updating. Also, the behavior between onNextClick and onPrevClick should be consolidated into a single block of logic. Current state of this with these changes are inconsistend - I'm not sure if this is still in progress though.

media/js/src/utils.js
```diff
@@ -814,6 +814,22 @@ const getTerms = function(annotations) {
...
+            'alt', `Annotation titled ${title}`);
+    } else if (type === 'audio') {
+        document.getElementById(`audio-${id}`).setAttribute(
+            'alt', `Annotation titled ${title}`);
+    } else if (type === 'unknown') {
+        document.getElementById(`unknown-${id}`).setAttribute(
+            'alt', `Annotation titled ${title}`);
+    } else {
+        document.getElementById(`map-${id}`).setAttribute(
+            'aria-label', `Annotation titled ${title}`);
```
nikolas, line 829: This way of updating things can work.. but this is a sort of imperative, "jQuery"-way of updating things, that is prone to breaking with future changes.

A more solid, declarative approach would be better. Since we happen to already be in React here, we can do that. The dom element's alt text can be declared more directly in the parent component - GridAsset.jsx.

The parent component already has a `selectedAnnotation` state variable, and we can use that to determine what the element's alt text should be in a more direct fashion. And also, any other attribute we want to set. Altering the elements after they've been rendered, with setAttribute, is unnecessary.

media/js/src/AnnotationScroller.jsx
ndittren: @nikolas definitely deleted by accident

The idea behind the logic is if that the alt label would change according to the annotation but at one point the on prev click would reach the original asset which would change the wording of the alt to identify it as an Asset and not an Annotation.

media/js/src/GridAsset.jsx
```diff
@@ -100,7 +101,15 @@ export default class GridAsset extends React.Component {
...
         const a = this.props.asset.annotations[annotation];
 
-        this.setState({selectedAnnotation: a});
+        this.setState({selectedAnnotation: a},
+            function(){
+                if (this.state.selectedAnnotation){
+                    this.setState({annotationLabel: `annotation: ${this.state.selectedAnnotation.title}`});
+                } else {
+                    this.setState({annotationLabel: `asset: ${this.props.asset.title}`});
+                }
```
nikolas, line 110: There are a few problems with this logic here.

First, the nested `this` on line 106 refers to the immediately surrounding function being defined at line 105. To get to the components `this`, you would make a new variable to refer to it, e.g. `const me = this;` in the outer context.

Second, we don't want to use `setState` in a chain like this - `setState` is meant to be used in a single pass, and calling it multiple times like this can have unintended side effects.

The logic here can be refactored for a single setState call, e.g. something like:
```
let label = `asset: ${this.props.asset.title}`;
if (a) {
    label = `annotation: ${a.title}`;
}

this.setState({
    selectedAnnotation: a,
    label: label
});
```

I renamed `annotationLabel` to just `label`, since it may refer to either the asset or the annotation.

nikolas:
Also, a side note - I've found merge conflicts easier to resolve when the updated `media/js/bundle.js` isn't included in the pull requests making the actual code changes.

media/js/src/GridAsset.jsx
```diff
@@ -99,8 +100,12 @@ export default class GridAsset extends React.Component {
     onSelectedAnnotationUpdate(annotation) {
         const type = this.asset.getType();
         const a = this.props.asset.annotations[annotation];
+        let label = `asset: ${this.props.asset.title}`;
 
-        this.setState({selectedAnnotation: a});
+        if (a) {
+            label = `annotation: ${a.title}`;
+        }
+        this.setState({selectedAnnotation: a, label: label});
```
nikolas, line 108: I just realized, this `label` state variable is unnecessary - it's just derived off the `selectedAnnotation` state. It's better to handle this in the `render` step, to simplify state changes. State variables should only be used when absolutely necessary. I'll add a commit here to do this refactor.

ndittren: Yeah I tried this originally and kept getting an error when I would click back to the original asset for some reason... Not doing it now though

PR #4675, opened 2022-10-14
switch to material-ui table - issue#4648

ndittren:
@nikolas got this mostly working. The feature missing is the sorting. But I am getting a React same key warning in the console that popped in when I made the assets column links. Tried adding a key but didn't fix the issue. This doesn't break anything though.

### DIFF
--- a/media/js/src/CollectionListView.jsx
+++ b/media/js/src/CollectionListView.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+
+/* material-ui includes */
 import { styled } from '@mui/material/styles';
 import Table from '@mui/material/Table';
 import TableBody from '@mui/material/TableBody';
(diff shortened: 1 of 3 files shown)


nikolas:
Dang, yeah I figured the sorting may be the tricky part. I will take a look. Also - see my note on the other PR: when making React changes it's easier to work with just the source file changes, not the compiled bundle.js stuff, so you can leave that out of this PR.

ndittren:
Sorting added

nikolas:
The sorting is actually handled on the backend because it sorts the course's entire asset collection, not just the current page. See: https://mediathread.ctl.columbia.edu/course/1186/

We should be able to adjust this to work in a similar way.

nikolas:
Also, we currently only have it sortable by Title, Owner, and Date, so I've made that adjustment in the code here.

nikolas:
Hah - so when looking at this sort stuff I came across mui's react-data-grid component which supports server-side sorting in a really straightforward way: https://mui.com/x/react-data-grid/sorting/#server-side-sorting

Unfortunately, this library renders everything as divs just like our current library. So, I will just continue here and see if I can move things forward using react-table.

nikolas:
Hmmm, and this table library may do everything we want it to do. It looks a little complicated to set up but I'm going to look into this on a different branch and see how it works:
* https://github.com/TanStack/table
* server-side sort example: https://tanstack.com/table/v8/docs/examples/react/sorting

nikolas:
So, after looking at this problem from a few angles, (I even have a branch [here](https://github.com/ccnmtl/mediathread/tree/data-table-semantic) that updates the existing library to use `<table>` and `<tr>` elements), I think our current table setup here may be okay. Through attempting to re-implement sorting with different libraries, I'm realizing this is a problem that merits a bigger timeframe than we currently have to work with for these changes.

Although the react-data-table-component table is using divs, ARIA roles are defined properly, so this should suffice for screen readers.

PR #4682, opened 2022-10-19
Added aria-labels in asset detail view

ndittren:
* Since this logic was set up in the collection space view, just moving it over to the asset detail space.

### DIFF
--- a/media/js/src/assetDetail/AssetDetail.jsx
+++ b/media/js/src/assetDetail/AssetDetail.jsx
@@ -712,6 +712,11 @@ export default class AssetDetail extends React.Component {
         let media = null;
 
         let invisibleEl = null;
+        let label = `asset: ${this.props.asset.title}`;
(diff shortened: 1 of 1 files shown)


media/js/src/assetDetail/AssetDetail.jsx
```diff
@@ -789,8 +794,9 @@ export default class AssetDetail extends React.Component {
                     {annotationTools}
                     <div
                         id={`map-${this.props.asset.id}`}
-                        className="ol-map"
-                        aria-label={`Asset titled ${this.props.asset.title}`}></div>
+                        className="ol-map target"
```
nikolas, line 797: Is the new `target` class doing anything here?

ndittren: Ah no. I used that for testing. Will remove.

PR #4703, opened 2022-10-26
Add Asset.transcript functionality - closes #4646

### LINKED ISSUE #4646: Enable transcript display for audio/video assets
The asset author should be able to upload a transcript to a media asset, displayed beside the media for all users.

nikolas:


media/js/src/assetDetail/TranscriptForm.jsx
```diff
@@ -0,0 +1,103 @@
...
+import PropTypes from 'prop-types';
+
+import {updateAsset, truncateString} from '../utils';
+
+export default class TranscriptForm extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            isEditing: false,
+            transcript: this.props.asset.transcript
```
ndittren, line 11: @nikolas the page is crashing when I click on View Metadata because this.props.asset.transcript is null

nikolas: oh weird - i'll take a look and see if I can reproduce that. And in any case I'll make the code here more forgiving and allow it to just be null if there's no asset loaded yet. I often have to do that when making react code that relies on an XHR request like this.

nikolas: Okay - yeah I reproduced the problem, and the crash actually starts in the truncateString() function. I made the fix - that's what I get for pushing this out with no unit tests..

PR #4715, opened 2022-11-04
⬆️ celery 5.2.7

ndittren:


requirements.txt
```diff
@@ -138,9 +138,17 @@ billiard==3.6.4.0
 vine==1.3.0
 amqp==2.6.1
 amqplib==1.0.2
```
nikolas, line 140: vine and amqp need some updates here as well. It looks like [amqplib](https://pypi.org/project/amqplib/#history) is from 2011, and not used in wardenclyffe. I'm pretty sure this can just be removed.

PR #5191, opened 2023-11-27
Revert s3transfer to 0.7.0

Evan-CTL:


ndittren:
Don't think this is necessary

PR #5354, opened 2024-04-18
:arrow_up: django 4.2

nikolas:


nikolas:
This is just failing on the intermittent cypress failures now, so I'm trying to get that to pass.

nikolas:
Looks like there may be a valid failure here - a CSRF error on Composition creation. I'm looking into it.

PR #5389, opened 2024-05-17
publishtoworld cypress test: minor adjustments

nikolas:


nikolas:
This test seems to be regularly failing with a React error, maybe a bug in the component rendering templates: https://react.dev/errors/31?invariant=31&args%5B%5D=%5Bobject%20Error%5D

> Objects are not valid as a React child (found: [object Error]). If you meant to render a collection of children, use an array instead.

nikolas:
Actually, i think this PR may address that react error, at least: https://github.com/ccnmtl/mediathread/pull/5391

I am seeing it locally in my browser on the project list page, but making a new build via `npm run build` fixes it.

nikolas:
Nevermind - getting that same error on this test even with the new build.

PR #5402, opened 2024-05-21
Test on python 3.12

nikolas:


nikolas:
I've added a custom-built pylibmc wheel file here which should support python 3.12.

PR #5404, opened 2024-05-21
Stringify error if it's an object

nikolas:
This may help mitigate this react error we're seeing: https://react.dev/errors/31?invariant=31&args%5B%5D=%5Bobject%20Error%5D

The React code expects 'e' to be a string, but that may not always be the case, so cast it if it's an object.


nikolas:
@ndittren I think I finally tracked down this react error one of the tests was triggering.. crossing my fingers. Also, this did require a new JS build via `npm run build`. I think this probably should be automated in the build process.. going forward.

PR #5860, opened 2025-07-02
Updated the contact email

Evan-CTL:


nikolas:
So, we don't have this documented well, but ctl-mediathread@columbia.edu actually is a group email which I know is working, and which we use. I don't have a good understanding of where this is managed to confirm all this, but I think we want to keep this email as ctl-mediathread.

nikolas:
Yeah - i'm seeing this in grouper now. We have this email set up to forward to me, Natalia, Michael Tarnow and Mark Phillipson.

